PR #5, opened 2020-06-14
Always insert newline after 'do'

### LINKED ISSUE #1: Fix the test suite
Right now, the test suite is expecting everything to have 2 space indentation.

Reformatting it to have 4 space indentation would be great! Then we could publish to `stackage`.

georgefst:
Ugh, this is potentially a one-line fix, except that a handful of tests now fail with idempotence issues...

I guess we've gotta ask how much we care about holding ourselves to the same standards as the main `ormolu` branch in that respect. The upshot of not caring here would be that users may sometimes have to format twice if they write some really odd code using unusually compact `do`-notation.

Idempotence is obviously a valuable property, but if we start allowing some configuration, it's going to be increasingly difficult to guarantee in all cases anyway.

Fixes #1 (for good this time)

### DIFF
--- a/data/examples/declaration/splice/bracket-declaration-out.hs
+++ b/data/examples/declaration/splice/bracket-declaration-out.hs
@@ -14,7 +14,8 @@ foo =
         deriving (Eq, Ord, Enum, Bounded, Show)
     |]
 
-$(do [d|baz = baz|])
+$(do
+      [d|baz = baz|])
 
 $(singletons [d|data T = T deriving (Eq, Ord, Enum, Bounded, Show)|])
 
--- a/data/examples/declaration/value/function/block-arguments-out.hs
+++ b/data/examples/declaration/value/function/block-arguments-out.hs
@@ -1,4 +1,5 @@
-f1 = foo do bar
+f1 = foo do
+    bar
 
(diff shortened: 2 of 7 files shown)


georgefst:
> Fixes #1 (for good this time)

Well, only if we disable the idempotence tests, obviously.

parsonsmatt:
What's the idempotency tests that are failing?

I'm in favor of this. I have occasionally used a form with blockarguments like

```haskell
map
  do \x -> ...
  do filter p xs
```

but I would not be upset if a newline happened.

georgefst:
Actually, they mostly come down to the fact that `ormolu` allows braces in single-line `do`s, but expands them in multi-line. Seeing as this patch makes all `do`s multi-line, we should be able to just make sure this transformation is always applied on the first pass. I'm currently unsure where the relevant code lives, however.

The one other case then is the `bracket-declaration` test, where we have:

``` haskell
$( do [d| baz = baz |] )
```

formatted to:

``` hs
$(do
      [d|baz = baz|])
```

then:

```hs
$( do
       [d|baz = baz|]
 )
```
...

georgefst:
Closing in favour of #6.

PR #9, opened 2020-06-18
Add a config file

### LINKED ISSUE #2: Add a configuration file
It should be possible to configure `fourmolu`.

A great initial choice would be the amount of space in the indentation.

I'm not picky on the file format. Dhall might be cool? Yaml is a safe bet.

georgefst:
Closes #2.

Fourmolu looks through parent directories, then [the XDG config directory](https://hackage.haskell.org/package/directory-1.3.6.1/docs/System-Directory.html#v:XdgConfig), for a `fourmolu.yaml`, which is used to override config fields. Right now, the only field is `indentation`.

So a config file just looks like:
```yaml
indentation: 4
```

### DIFF
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -10,7 +10,6 @@ import Control.Exception (SomeException, displayException, try)
 import Control.Monad
 import Data.Either (lefts)
 import Data.List (intercalate, sort)
-import Data.Maybe (fromMaybe)
 import qualified Data.Text.IO as TIO
 import Data.Version (showVersion)
 import Development.GitRev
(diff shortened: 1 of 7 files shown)


georgefst:
Hmm, not quite sure what happened with the commit history there - I think I may have accidentally rebased twice. Anyway, the end result is as intended.

georgefst:
Actually, this isn't quite right yet. Using `getCurrentDirectory` to start searching for config files isn't always the right approach, e.g. when being called from Haskell Language Server.

georgefst:
Closing in favour of #10, which allows a directory other than `getCurrentDirectory` to be specified. Seemed worth making a fresh PR due to the messy commit history.

PR #10, opened 2020-06-22
Add support for config files

### LINKED ISSUE #2: Add a configuration file

georgefst:
Closes #2.

Fourmolu looks through parent directories, then [the XDG config directory](https://hackage.haskell.org/package/directory-1.3.6.1/docs/System-Directory.html#v:XdgConfig), for a `fourmolu.yaml`, which is used to override config fields. Right now, the only field is `indentation`.

So a config file just looks like:
```yaml
indentation: 4
```

### DIFF
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -183,6 +183,7 @@ configParser =
                 help "End line of the region to format (inclusive)"
               ]
         )
+    <*> pure defaultPrinterOpts
 
 ----------------------------------------------------------------------------
 -- Helpers
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -25,7 +25,10 @@ import System.IO (hPutStrLn, stderr)
 main :: IO ()
 main = withPrettyOrmoluExceptions $ do
   Opts {..} <- execParser optsParserInfo
-  let formatOne' = formatOne optMode optConfig
+  let formatOne' path = do
+        printerOpts <-
+          loadConfigFile (cfgDebug optConfig) path $ cfgPrinterOpts optConfig
+        formatOne optMode optConfig {cfgPrinterOpts = printerOpts} path
   case optInputFiles of
     [] -> formatOne' Nothing
     ["-"] -> formatOne' Nothing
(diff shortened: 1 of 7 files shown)


fourmolu.yaml
```diff
@@ -0,0 +1 @@
+indentation: 2
```
parsonsmatt, line 1: hey :triumph:

app/Main.hs
```diff
@@ -183,6 +186,7 @@ configParser =
                 help "End line of the region to format (inclusive)"
               ]
         )
+    <*> pure defaultPrinterOpts
```
parsonsmatt, line 189: hmmmm

it feels to me like we should allow the printer options to be configurable on the command line as well

Not a blocking comment, we can make an issue for that.

app/Main.hs
```diff
@@ -25,7 +25,10 @@ import System.IO (hPutStrLn, stderr)
 main :: IO ()
 main = withPrettyOrmoluExceptions $ do
   Opts {..} <- execParser optsParserInfo
-  let formatOne' = formatOne optMode optConfig
+  let formatOne' path = do
+        printerOpts <-
+          loadConfigFile (cfgDebug optConfig) path $ cfgPrinterOpts optConfig
+        formatOne optMode optConfig {cfgPrinterOpts = printerOpts} path
```
parsonsmatt, line 31: Note that when/if it becomes possible to configure the printer options via the CLI, we'll have to have a merge operation here.

fourmolu.yaml
georgefst: Ha, yeah, it does seem counter-intuitive.

But this way we can keep the diffs simpler when looking at upstream changes. Although I suppose it is just whitespace...

app/Main.hs
georgefst: Oh yeah, it would definitely be worthwhile - I just decided to keep this minimal.

(My contributions so far have mostly been based on what I personally need to be productive, and since I'm now primarily using Fourmolu via (a fork of) HLS, anything CLI-related isn't much of a priority)

PR #15, opened 2020-07-17
Add --indentation CLI option.

kukimik:
This took me longer than I planned, but finally here's a working version that solves #12. In the end I decided to follow the approach proposed by @georgefst, but perhaps I've stopped a few steps earlier than it was intended.

I've had a number of decisions to make, and I'll try to explain myself.

* There's no short version of the `--indentation` option: I wanted to avoid conflicts with `ormolu` if they add an `-i` option in the future.
* I'm not sure if deriving `Eq` and `Show` for `PrinterOptsPartial` and `PrinterOptsTotal` is done the "right" way. I wanted to stay conservative. The  `StandaloneDeriving` extension seemed safer to me than `UndecidableInstances`, which could perhaps allow me to save a few lines (but I'm not even sure about that, I have not tried).
* I've not defined a HKD type family, as described in some tutorials like https://reasonablypolymorphic.com/blog/higher-kinded-data/, which would allow me to get rid of the `Identity` wrapping. I don't find the `Identity` such a great nuisance at the moment. I don't understand how the type families work, so I didn't want to introduce "magic" (and an additional language extension) just to clean it up.
* I have not defined a `Semigroup` instance for `PrinterOptsPartial`. I don't really need it at the moment, and I didn't know how to do it in a slick way. If I wrote an instance by hand, then every time a new field was added to `PrinterOpts` one would have to change both the `Semigroup` instance and another function, used to merge `PrinterOptsPartial` with `PrinterOptsTotal`. I don't know how to do it in a generic way without much effort or additional dependencies.
* On the other hand I'm not really satisfied with the `mergePrinterOpts` function:
```haskell
mergePrinterOpts :: [PrinterOptsPartial] -> PrinterOptsTotal -> PrinterOptsTotal
mergePrinterOpts opts totalOpts = PrinterOpts{ poIndentation = m poIndentation poIndentation }
  where
    m :: (PrinterOptsTotal -> Identity a) -> (PrinterOptsPartial -> Maybe a) -> Identity a
    m ft fp = Identity $ fromMaybe (runIdentity . ft $ totalOpts) (asum $ map fp opts)
```
  I'd like to have something like `m :: (PrinterOpts f -> f a) -> Identity a`, so that I could write `poIndentation = m poIndentation`, but again I don't see how this could be achieved without much effort.

### DIFF
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -33,8 +33,8 @@ main = withPrettyOrmoluExceptions $ do
           optConfig
             { cfgPrinterOpts =
                 mergePrinterOpts
-                  [optPrinterOpts, filePrinterOpts]
-                  defaultPrinterOpts
+                  (optPrinterOpts <> filePrinterOpts)
+                  (cfgPrinterOpts optConfig)
             }
           path
   case optInputFiles of
@@ -197,7 +197,7 @@ configParser =
                 help "End line of the region to format (inclusive)"
               ]
         )
-    <*> pure defaultPrinterOpts -- just a placeholder that is overwritten later
+    <*> pure defaultPrinterOpts
 
 printerOptsParser :: Parser PrinterOptsPartial
 printerOptsParser =
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -26,9 +26,17 @@ main :: IO ()
 main = withPrettyOrmoluExceptions $ do
   Opts {..} <- execParser optsParserInfo
   let formatOne' path = do
-        printerOpts <-
-          loadConfigFile (cfgDebug optConfig) path $ cfgPrinterOpts optConfig
-        formatOne optMode optConfig {cfgPrinterOpts = printerOpts} path
+        filePrinterOpts <-
+          loadConfigFile (cfgDebug optConfig) path
+        formatOne
+          optMode
+          optConfig
+            { cfgPrinterOpts =
+                mergePrinterOpts
+                  [optPrinterOpts, filePrinterOpts]
+                  defaultPrinterOpts
+            }
+          path
   case optInputFiles of
     [] -> formatOne' Nothing
     ["-"] -> formatOne' Nothing
@@ -85,6 +93,8 @@ data Opts = Opts
     optMode :: !Mode,
     -- | Ormolu 'Config'
     optConfig :: !(Config RegionIndices),
+    -- | Fourmolu-specific options
+    optPrinterOpts :: !PrinterOptsPartial,
     -- | Haskell source files to format or stdin (when the list is empty)
     optInputFiles :: ![FilePath]
   }
@@ -145,6 +155,7 @@ optsParser =
         help "Mode of operation: 'stdout' (default), 'inplace', or 'check'"
       ]
     <*> configParser
+    <*> printerOptsParser
     <*> (many . strArgument . mconcat)
       [ metavar "FILE",
         help "Haskell source files to format or stdin (default)"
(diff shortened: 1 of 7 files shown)


kukimik:
On a second thought, it's really better to have `mergePrinterOpts` and `completePrinterOpts` as separate functions. I'll fix that.

kukimik:
We now still have a single function, but one that specializes to the `mergePrinterOpts` and `completePrinterOpts` that you proposed. I think that this combines the advantages of both of the approaches.

app/Main.hs
```diff
@@ -188,6 +199,15 @@ configParser =
         )
     <*> pure defaultPrinterOpts
 
+printerOptsParser :: Parser PrinterOptsPartial
+printerOptsParser =
+  PrinterOpts
+    <$> (optional . option auto . mconcat)
+      [ long "indentation",
+        metavar "WIDTH",
+        help "Number of spaces per indentation step"
```
georgefst, line 208: For consistency, I'd put the default value in brackets in the help text.

Not super necessary I suppose, as it is in the name...

src/Ormolu/Config.hs
```diff
@@ -93,15 +99,50 @@ defaultConfig =
...
+defaultPrinterOpts :: PrinterOptsTotal
+defaultPrinterOpts = PrinterOpts {poIndentation = Identity 4}
+
+-- | Fill the field values that are 'Nothing' in the first argument
+-- with the values of the corresponding fields of the second argument.
+mergePrinterOpts ::
+  (Applicative f) =>
+  PrinterOptsPartial ->
+  PrinterOpts f ->
+  PrinterOpts f
```
georgefst, line 139: I'd rename this to something like `completePrinterOpts`, since to me "merge" implies something that would take two arguments of the same type.

Perhaps that's not quite ideal either, given that it's ambiguous at a glance whether "complete" is being used as a verb or a noun...

src/Ormolu/Config.hs
```diff
@@ -93,15 +99,50 @@ defaultConfig =
...
+mergePrinterOpts p1 p2 =
+  PrinterOpts {poIndentation = m poIndentation poIndentation}
+  where
+    m f1 f2 = case f1 p1 of
+      Nothing -> f2 p2
+      Just x -> pure x
```
georgefst, line 145: I think this would be clearer with `m` renamed to `mergeField` (or `completeField`). Or with the type signature added.

georgefst:
> There's no short version of the --indentation option

Right call. They've actually added `-i` in the last few weeks as a shorthand for `--mode inplace`.

> I'm not sure if deriving Eq and Show for PrinterOptsPartial and PrinterOptsTotal is done the "right" way

I think the `UndecidableInstances` approach would be worth thinking about if we expected to ever want the parameter to be anything other than `Identity` or `Maybe`, but I don't really see that happening. Current approach is certainly simpler, and that's no bad thing.

georgefst:
Apologies for taking so long to get around to this...

src/Ormolu/Config.hs
georgefst: "fill", "saturate" and "augment" are the other alternatives I can think of. This is obviously not worth bikeshedding too far.

georgefst:
(Ignore that. I accidentally committed to the wrong branch. We're back to where we were.)

src/Ormolu/Config.hs
kukimik: I've renamed it to `fillMissingPrinterOpts`.

src/Ormolu/Config.hs
kukimik: Renamed.

app/Main.hs
kukimik: I've added "(default 4)" to the help text. But this is not entirely true, as the value is read from the config file if not provided in the CLI option, and only if it is not in the file, it defaults to 4. However, my I've no idea how to express this in just a few words.

kukimik:
Thanks for the review. I didn't know what is the conversation resolving policy in this project (at work we leave it to the reviewer), so I've kept them open.

I'm away for a week, so if there are any other comments, they'll have to wait...

app/Main.hs
georgefst: That's a good point.

I think this'll do for now. Once we have more configuration, what we probably want is to have a separate section in the help text, corresponding to `PrinterOpts`. Then we can make clear at the top of that section that these same settings can be set via config file. I think `optparse-applicative` makes this fairly easy, but I can't remember how it works off the top of my head.

georgefst:
> I didn't know what is the conversation resolving policy in this project

Personally I'm happy for anyone to resolve when there's obviously no issue, as with the renamings here.

PR #17, opened 2020-08-07
Add option for leading commas (and expand test suite)

### LINKED ISSUE #16: Commas first
Since we have opened the pandoras box of configuration, would it be possible to add an option for having commas first in lists and records?

It is the one thing that annoys me whenever I use ormolu and it is the opposite of how all the rest of the Haskell community does it<sup>e.g. [1](https://kowainik.github.io/posts/2019-02-06-style-guide) [2](https://github.com/tibbe/haskell-style-guide/blob/master/haskell-style.md)</sup>. The motivation seems fairly lacking, since Haskell afaik doesn't support trailing commas in most cases and now it forces you to manually type commas when appending a line to a list.

How tightly integrated is the choice of commas last in the design of the program? Would adding this configuration lead to a lot of extra complexity?

georgefst:
Closes #16 

Taken almost directly from [the google fork](https://github.com/google/ormolu/commit/5a36b8b6ef85b587bbf6e8cd5ecb7754fed7461d).

This also expands the test suite so that we test both with our default options, and with Ormolu's. Currently, that covers all options, though not all combinations. In hindsight, I should have implemented this as a separate change first. But it didn't seem worth splitting up after the fact.

This doesn't implement leading commas for imports and exports, since the implementation there is more complex, and I believe they should be controlled by a separate option anyway, not least because they are allowed to have an extra comma. I'm hoping to implement that option this weekend.


src/Ormolu/Config.hs
```diff
@@ -128,22 +132,40 @@ deriving instance Eq PrinterOptsTotal
 deriving instance Show PrinterOptsTotal
 
 defaultPrinterOpts :: PrinterOptsTotal
-defaultPrinterOpts = PrinterOpts {poIndentation = Identity 4}
+defaultPrinterOpts =
+  PrinterOpts
+    { poIndentation = pure 4,
+      poCommaStyle = pure Leading
```
georgefst, line 138: Whether we want this by default is up for debate.

PR #24, opened 2020-09-20
Fix formatting script

georgefst:
I got burnt by the fact that it was using the globally installed `fourmolu`, rather than the current source...

This fix still isn't really ideal, as there can be a lot of recompilation before reaching a fixed point. What might be nicest is to wait until `cabal run` allows using the bytecode interpreter.

### DIFF
--- a/format.sh
+++ b/format.sh
@@ -4,6 +4,7 @@ set -e
 
 cabal format
 
-cabal run fourmolu -- -i $(find app -type f -name "*.hs")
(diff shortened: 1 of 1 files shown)


georgefst:
Whoops - make sure to squash this.

format.sh
```diff
@@ -2,10 +2,9 @@
 
 set -e
 
-export LANG="C.UTF-8"
```
georgefst, line 5: Unrelated change really, but this was causing issues on my machine, and I haven't really debugged why...

Issue #19, opened 2020-08-07
Ormolu rejects


georgefst:
A list of features rejected by Tweag (not quite definitively in some cases) that we might want to have as options:

- [ ] Prettier and more diff-friendly import/export lists: tweag/ormolu#409
- [ ] Respect more newlines: tweag/ormolu#74 (parts of this have been implemented in Ormolu)
  - [ ] Import groups
    - It's worth thinking about how this plays along with tooling for auto-inserting imports e.g. HLS
  - [ ] Not always one line between top-level definitions
    - Multiple blank lines can be useful for semantic grouping
    - Zero is nice for trivial definitions, especially instances
- [ ] Break up long lines: tweag/ormolu#508
  - It's possible that this contradicts Ormolu's design enough to be a nightmare to implement fully
- [ ] No space before record opening brace: tweag/ormolu#565

ProofOfKeags:
I just opened #45 as an explicit issue regarding import stanzas. HLS will insert auto imports immediately under the last import statement in your module. If there is a common pattern around stanzas we could encode the preferences in pragmas such that the total set of imports could uniquely determine the import grouping.

YellowOnion:
Stylish Haskell has qualified import's aligned column wise, any chance to have it look like this?

```hs
import           Control.Concurrent       (getNumCapabilities)
import           Control.Concurrent.Async
import           Control.Concurrent.MVar
import           Control.Monad            (forM_, when)
import           GHC.Clock
import qualified System.Posix             as Unix
```

brandonchinn178:
Feel free to open a new issue! Personally, I don't know why anyone would prefer that format, especially with ImportQualifiedPost, but if people are interested, feel free to submit a PR

PR #26, opened 2020-09-20
Add option to not insert a space before record braces

georgefst:
Closes one of #19.


georgefst:
I opted to keep the space in data type declarations. It just feels weird somehow:

```hs
data Foo = Foo {fooX :: Int, fooY :: Int}
data Foo = Foo{fooX :: Int, fooY :: Int}
```

Anyway, in general, the major reason for preferring not to have the space, is to underline the fact that record update/construction binds even more tightly than function composition, and that isn't relevant here.

I'm open to changing it if anyone has an opinion.

PR #28, opened 2020-09-21
Better import/export lists

georgefst:


src/Ormolu/Printer/Meat/Module.hs
```diff
@@ -56,16 +58,21 @@ p_hsModule mstackHeader shebangs pragmas qualifiedPost HsModule {..} = do
...
         forM_ hsmodDeprecMessage $ \w -> do
-          located' p_moduleWarning w
           breakpoint
+          located' p_moduleWarning w
+        breakIfNotDiffFriendly
+
+        -- This works around an awkward idempotency bug with deprecation messages.
+        diffFriendly <- getPrinterOpt poDiffFriendlyImportExport
+        when (diffFriendly && not (null hsmodDeprecMessage)) newline
+
```
georgefst, line 69: This is annoying, and the end result is not very pretty (see the examples containing "warning" in the diff). But I couldn't find a better solution, and I'd already spent far too much time thinking about this edge case.

georgefst:
Closes another of #19.

src/Ormolu/Printer/Meat/ImportExport.hs
```diff
@@ -135,11 +138,27 @@ commaDel' :: R ()
...
+    body
+    txt ")"
   where
+    body = vlayout singleLine multiLine
     singleLine = m
     multiLine = do
       space
       sitcc m
       newline
+    trailingParen = if topLevelImport then txt " )" else txt ")"
```
georgefst, line 159: The complication is that we can't have a trailing parenthesis in the first column for import lists, i.e.

```hs
import M (
    x,
)
```

As you can see, my preferred solution is just to add a single space, but this may not be the most popular.

georgefst, line 159: The complication here is that we can't have a trailing parenthesis in the first column, for import lists, i.e.

```hs
import M (
    x,
)
```

As you can see, my preferred solution is to add a single space, but this may not be the most popular.

georgefst:
@parsonsmatt This is probably the last of this round of features. Would be great if you could lend a quick glance.

PS. Have you started actually using Fourmolu yet? I'm trying it on some smaller projects and it's mostly quite nice now, but I'm still wanting it to be a bit less opinionated about blank lines. I hoped that might change upstream, but it's looking less likely, so I guess that's the next thing I'll be looking at.

georgefst:
> This is probably the last of this round of features.

I lied: #29

steshaw:
@georgefst, thanks for implementing https://github.com/tweag/ormolu/issues/409! I am a Fourmolu convert!

PR #29, opened 2020-09-21
Preserve newlines

georgefst:


georgefst:
This keeps blank lines between import groups, and keeps top-level definitions together when there were originally no blank lines between them.

We can reuse this flag for any similar changes, but these two seem to be the major issues.

PR #30, opened 2020-09-23
Use multiline comment style for multiline haddocks

georgefst:
My initial intention here was just to preserve haddock comments completely. But that seems to be difficult, since the GHC API doesn't actually tell us, for example, whether a comment used single-line (`--`) or multi-line (`{- -}`) style. We *could* use the `SrcSpan`s to grab the appropriate text from the input verbatim, but that's a bit nasty.

So, I've gone with @michaelpj's suggestion from [the Ormolu thread about this](https://github.com/tweag/ormolu/issues/641), which is to always print in multi-line style for multi-line haddocks, even if they originally just used multiple `--`s. One slight gotcha is that:

```hs
-- | Multi
-- line
-- comment
```

becomes:
```hs
{- | Multi
 line
 comment
-}
```

, when it might seem more natural for the leading spaces to be removed. And, depending on the user's editor, this might be non-trivial to "fix". But I can't see a good general solution to that, that wouldn't strip leading whitespace from multi-line-style comments, where it is likely to be intentional.

If we care about keeping the guarantee that Fourmolu is completely compatible with Ormolu (e.g. so we can continue to dogfood without introducing large diffs vs. upstream), then we ought to make this configurable like everything else. Although I do find the existing behaviour difficult to justify.

### DIFF
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -244,6 +244,11 @@ printerOptsParser =
         metavar "BOOL",
         help "Give the programmer more choice on where to insert blank lines (default 'true')"
       ]
+    <*> (optional . option parseHaddockStyle . mconcat)
+      [ long "haddock-style",
+        metavar "STYLE",
+        help "How to print Haddock comments (default 'multi-line')"
+      ]
 
 ----------------------------------------------------------------------------
 -- Helpers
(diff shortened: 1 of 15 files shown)


georgefst:
This would close another of #19.

michaelpj:
I wonder whether it would be worth opening a PR with the "opinionated" version of this (i.e. always make comments multiline) to `ormolu`. I don't feel that the discussion was conclusive, they might actually respond to an implementation. Might be a waste of time, though.

(If you're not keen on doing that, do you mind if I port your patch over and give it a shot?)

georgefst:
I've sensed from Mark's recent comments that they're pretty reluctant to change anything now. But admittedly that issue hasn't been closed so it's not impossible in this case. I won't be pursuing it, but feel free to go ahead - Fourmolu really hasn't diverged much and doesn't intend to, so it should just lift straight out.

PR #36, opened 2020-09-29
Fix HsYAML-aeson revision to use compatible aeson

### LINKED ISSUE #35: fourmolu fails to build with current stack.yaml
In a clean environemnt (the `haskell:8.8.3` image from dockerhub), running `stack install` with no modifications leads to this:
```
Error: While constructing the build plan, the following exceptions were encountered:
 In the dependencies for HsYAML-aeson-0.2.0.0:
     aeson-1.5.2.0 from stack configuration does not match ^>=1.4.0.0  (latest matching version
                   is 1.4.7.1)
 needed due to fourmolu-0.1.0.0 -> HsYAML-aeson-0.2.0.0
 Some different approaches to resolving this:
   * Set 'allow-newer: true'
     in /root/.stack/config.yaml to ignore all version constraints and build anyway.
   * Recommended action: try adding the following to your extra-deps in /fourmolu/stack.yaml:
 - aeson-1.4.7.1@sha256:6d8d2fd959b7122a1df9389cf4eca30420a053d67289f92cdc0dbc0dab3530ba,7098
 Plan construction failed.
```
Looking at the `.cabal` for `HsYaml-aeson-0.2.0.0` this is [indeed true](http://hackage.haskell.org/package/HsYAML-aeson-0.2.0.0/src/HsYAML-aeson.cabal). Interestingly, if I build once with `allow-newer: true` and then remove it I can still build afterwards (indicating some kind of caching bug in stack).

In any case, what would you like to do to fix this?
* use `allow-newer: true`
* use [`r2`](http://hackage.haskell.org/package/HsYAML-aeson-0.2.0.0/revisions/) for `HsYAML-aeson` which fixes this bound

googleson78:
Fixes #35

### DIFF
--- a/stack.yaml
+++ b/stack.yaml
@@ -42,6 +42,7 @@ packages:
 # extra-deps: []
 
 extra-deps:
+  - HsYAML-aeson-0.2.0.0@rev:2
   - aeson-1.5.2.0
 
 # Override default flag values for local packages and extra-deps


georgefst:
Thanks, my bad for letting the stack config slip (again...).

I had been meaning to double check that it's working before making a release (which should be very soon - I've got one remaining patch cleaning up nested lists etc.).

PR #37, opened 2020-10-01
Fix nested lists etc.

georgefst:
My third attempt at fighting all the subtle ways in which Ormolu encodes a preference for two-space indentation. Still room for improvement.

As ever, see the test diffs to get a sense of the effects.

This fixes my major remaining annoyance with Fourmolu.


src/Ormolu/Printer/Internal.hs
```diff
@@ -387,12 +387,14 @@ inci = inciBy 1
 -- | Like 'inci', but indents by the given fraction of a full step.
 inciBy :: Int -> R () -> R ()
 inciBy x (R m) = do
-  indentStep <- R (asks (runIdentity . poIndentation . rcPrinterOpts))
+  step <- (`quot` x) <$> R (asks (runIdentity . poIndentation . rcPrinterOpts))
   let modRC rc =
         rc
-          { rcIndent = rcIndent rc + indentStep `quot` x
+          { rcIndent = roundDownToNearest step (rcIndent rc) + step
```
georgefst, line 393: This is really fighting the symptoms rather than the cause.

But, well, it's a lot easier...

src/Ormolu/Printer/Meat/Declaration/Value.hs
```diff
@@ -1382,3 +1382,22 @@ breakpointPreRecordBrace = do
   if useSpace
     then breakpoint
     else breakpoint'
+
+listLike :: HsExpr p -> Bool
+listLike = \case
+  ExplicitList {} -> True
+  ExplicitTuple{} -> True
+  _ -> False
```
georgefst, line 1390: Alternative would just be to just look at whether the expression is multiline. But then we get occasional probably-unwanted newlines like:

```hs
[
    \x ->
    x
]

**Comment by georgefst** [comment]
The rounding causes some very minor changes to even the test outputs for 2 spaces. But they are arguably improvements:

```hs
f =
  False
    || ( 2
           >= 1
       )
```

Is it more important to have the `<=` operator in the column it would have been in without the line break? Or in a column that's actually a multiple of 2?

PR #42, opened 2020-10-02
Write loaded config info to stderr instead of stdout 

ollef:
Otherwise this becomes the first line of output when in stdout mode.

An alternative would be to use stderr (perhaps only when in stdout mode), if it's important information is shown prominently.


georgefst:
Ah, good catch - I think I've only ever used the inplace mode (`-i`), or editor integration, but I should've thought of this.

I would like this to be displayed prominently, so you could you possibly change this (and, less importantly, line 306) to always print to stderr?

ollef:
Done.

PR #44, opened 2020-10-06
Replace hardcoded defaults. Remove parser boilerplate. (#27)

### LINKED ISSUE #27: Make sure defaults don't slip out of sync
In `Main.printerOptsParser`, the default values are written as string constants, rather than coming from `defaultPrinterOpts`. We could easily forget to update these if we ever change defaults.

This is easy enough but should probably wait until after PRs adding options have been merged (#23, #26 etc.).

kukimik:
Closes #27.

This solves the problem described in #27. At the same time I wanted to reduce the number of places where the strings encoding option values were hardcoded, which resulted in rewriting/replacing `parseHaddockStyle`, `parseCommaStyle` and `parseBool` and modifying the way that the help texts are built.

I also decided to make the error messages a bit more user friendly by providing the user with the list of possible values of an option (because it was nearly free with the new approach).

Some comments:
1. I don't understand why booleans are case-insensitive and other options are case-sensitive. Do we really want it this way?
2. The values can still be represented in different ways on the command line and in the config file. I think it would be nice if we had a single source of truth, guaranteeing that these representations are uniform. However, I'm not sure what is the best approach to this problem. But anyway, I think this is out of scope for this issue/PR.

### DIFF
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -1,7 +1,10 @@
+{-# LANGUAGE AllowAmbiguousTypes #-}
 {-# LANGUAGE ApplicativeDo #-}
 {-# LANGUAGE CPP #-}
 {-# LANGUAGE LambdaCase #-}
+{-# LANGUAGE RankNTypes #-}
 {-# LANGUAGE RecordWildCards #-}
+{-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE TemplateHaskell #-}
 {-# LANGUAGE TypeApplications #-}
 
@@ -10,8 +13,8 @@ module Main (main) where
 import Control.Exception (SomeException, displayException, try)
 import Control.Monad
 import Data.Bool (bool)
-import Data.Char (toLower)
 import Data.Either (lefts)
+import Data.Functor.Identity (Identity (..))
 import Data.List (intercalate, sort)
 import Data.Maybe (fromMaybe)
 import qualified Data.Text.IO as TIO
@@ -114,7 +117,7 @@ data Mode
   | -- | Exit with non-zero status code if
     -- source is not already formatted
     Check
-  deriving (Eq, Show)
+  deriving (Eq, Show, Bounded, Enum)
 
 optsParserInfo :: ParserInfo Opts
 optsParserInfo =
@@ -157,7 +160,7 @@ optsParser =
             [ short 'i',
               help "A shortcut for --mode inplace"
             ]
-            <|> (option parseMode . mconcat)
+            <|> (option parseBoundedEnum . mconcat)
               [ long "mode",
                 short 'm',
                 metavar "MODE",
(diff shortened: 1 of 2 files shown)


georgefst:
This looks really good! Unfortunately I don't quite have time to look at the code in-depth right now, but I should be able to review properly by the weekend.

> I don't understand why booleans are case-insensitive and other options are case-sensitive. Do we really want it this way?

That's an error really. We probably want universal case-insensitivity, although in practice just accepting all-lowercase would be fine since this is what the examples use.

> The values can still be represented in different ways on the command line and in the config file. I think it would be nice if we had a single source of truth, guaranteeing that these representations are uniform. However, I'm not sure what is the best approach to this problem. But anyway, I think this is out of scope for this issue/PR.

Yeah, I'm interested in a general solution to this, and I think it could be quite neatly encapsulated in a library. See [my comment on Reddit](https://www.reddit.com/r/haskell/comments/ik1yp1/monthly_hask_anything_september_2020/), which was largely inspired by Fourmolu. It seems that nothing complete currently exists for Haskell. I've been meaning to research what solutions are available in other languages.

kukimik:
> We probably want universal case-insensitivity, although in practice just accepting all-lowercase would be fine since this is what the examples use.

I opt for requiring all-lowercase, to keep it compatible with Ormolu (and save a few lines of code).

When reviewing please take a look at `parseMode` in `Main.hs`. I left it untouched, as it comes from Ormolu and I wanted to save some merging work if they ever decide to change it. But perhaps it's better to make it uniform with other parsers.

> Yeah, I'm interested in a general solution to this, and I think it could be quite neatly encapsulated in a library. See my comment on Reddit, which was largely inspired by Fourmolu. It seems that nothing complete currently exists for Haskell. I've been meaning to research what solutions are available in other languages.

Sounds interesting.

georgefst:
> I opt for requiring all-lowercase, to keep it compatible with Ormolu (and save a few lines of code).

Yep, those are good enough reasons.

> When reviewing please take a look at `parseMode` in `Main.hs`. I left it untouched, as it comes from Ormolu and I wanted to save some merging work if they ever decide to change it. But perhaps it's better to make it uniform with other parsers.

I'd go for uniformity here. It's a trivial enough function that I'm not too concerned about merge conflicts. Would there be any visible change in behaviour?

> Sounds interesting.

Yeah, unfortunately, I'm really not looking for another library to maintain right now. Some of the code you've written here would actually be a good starting point, since it's really pretty independent of Fourmolu.

georgefst:
> I've thought about it for a moment (not a very long one wink ) and I don't know how to apply these ideas when we don't know what is the type that contains the preferences (i.e. when PrinterOpts f is replaced by an arbitrary type) and at the same time keep it easy to use.

Hmm, good point. I suppose we want to be able to work with any record type that has the correct kind ((`Type` -> `Type`) -> `Type`), with all fields being of the form `f a`, where a is an instance of the equivalent of `ToCLIArgument`. This is probably a bit much for GHC.Generics. We could use Template Haskell to generate some pieces.

> About the solutions in other languages: are you aware of some libraries that handle this?

I don't know any, but I feel like there must be something out there. I can't imagine we're the first to think about generalising this.

kukimik:
I did a quick research and found some libraries in:
* Clojure 
  * https://github.com/grammarly/omniconf
  * https://github.com/yogthos/config
* Go
  * https://github.com/spf13/viper
* Python
  * https://pypi.org/project/configglue/
  * https://github.com/bw2/ConfigArgParse
  * plus a StackOverflow question https://stackoverflow.com/questions/6133517/parse-config-files-environment-and-command-line-arguments-to-get-a-single-col

Didn't have time to read any of these, though.

This looks like an useful tool, and the problem seems rather accessible. If a project was started, I'd be eager to help. However, with my current Haskell skills, I don't think I'm ready to design a solution.

georgefst:
Nice. that list is very handy!

I'll let you know if I do make a start on this. Unfortunately, the likely need for TemplateHaskell to get something nice takes it out of the realms of "something I might be able to knock together on a Sunday afternoon".

PR #48, opened 2020-10-09
Configurable newlines between declarations

### LINKED ISSUE #47: Configurable number of lines between top-level definitions
We use 2 newlines between top-level definitions in our codebase. eg.

```haskell
f :: Int -> Int -> Int
f x y = square x + square y
  where
    square a = a * a


g :: Int -> Int
g z = f z z
```
as opposed to (single newline):
```haskell
f :: Int -> Int -> Int
f x y = square x + square y
  where
    square a = a * a

g :: Int -> Int
g z = f z z
```

It would be great if fourmolu had a configurable number of newlines between top-level definitions, such that we could set it to 2 instead of 1!

wraithm:
The number of newlines between declarations can now be configured with 'newlines-between-decls' configuration option.

Closes #47 

This does exactly what I want! It's idempotent as well. It defaults to exactly ormolu behavior. I tested it on our large industrial codebase. I don't see anything unexpected from my initial skim. If we merge this, I think we'd start using this at @bitnomial.

### DIFF
--- a/src/Ormolu/Printer/Internal.hs
+++ b/src/Ormolu/Printer/Internal.hs
@@ -18,6 +18,7 @@ module Ormolu.Printer.Internal
     atom,
     space,
     newline,
+    declNewline,
     useRecordDot,
     inci,
     inciBy,
@@ -316,6 +317,24 @@ space = R . modify $ \sc ->
         other -> other
     }
 
+declNewline :: R ()
+declNewline = newlineRawN . pred =<< getPrinterOpt poNewlinesBetweenDecls
+
+-- | Low-level newline primitive. This always inserts 'n' newlines.
+newlineRawN :: Int -> R ()
+newlineRawN n = R . modify $ \sc ->
+  sc
+    { scBuilder = scBuilder sc <> mconcat (replicate n "\n"),
+      scColumn = 0,
+      scIndent = 0,
+      scThisLineSpans = [],
+      scRequestedDelimiter = case scRequestedDelimiter sc of
+        AfterNewline -> RequestedNewline
+        RequestedNewline -> RequestedNewline
+        VeryBeginning -> VeryBeginning
+        _ -> AfterNewline
+    }
+
 -- | Output a newline. First time 'newline' is used after some non-'newline'
 -- output it gets inserted immediately. Second use of 'newline' does not
 -- output anything but makes sure that the next non-white space output will
@@ -365,7 +384,7 @@ newlineRaw = R . modify $ \sc ->
           scColumn = 0,
           scIndent = 0,
           scThisLineSpans = [],
-          scRequestedDelimiter = case scRequestedDelimiter sc of
+          scRequestedDelimiter = case requestedDel of
             AfterNewline -> RequestedNewline
             RequestedNewline -> RequestedNewline
             VeryBeginning -> VeryBeginning
--- a/src/Ormolu/Printer/Internal.hs
+++ b/src/Ormolu/Printer/Internal.hs
@@ -384,7 +384,7 @@ newlineRaw = R . modify $ \sc ->
           scColumn = 0,
           scIndent = 0,
           scThisLineSpans = [],
-          scRequestedDelimiter = case requestedDel of
+          scRequestedDelimiter = case scRequestedDelimiter sc of
             AfterNewline -> RequestedNewline
             RequestedNewline -> RequestedNewline
             VeryBeginning -> VeryBeginning
--- a/src/Ormolu/Printer/Internal.hs
+++ b/src/Ormolu/Printer/Internal.hs
@@ -318,22 +318,7 @@ space = R . modify $ \sc ->
     }
 
 declNewline :: R ()
-declNewline = newlineRawN . pred =<< getPrinterOpt poNewlinesBetweenDecls
-
--- | Low-level newline primitive. This always inserts 'n' newlines.
-newlineRawN :: Int -> R ()
-newlineRawN n = R . modify $ \sc ->
-  sc
-    { scBuilder = scBuilder sc <> mconcat (replicate n "\n"),
-      scColumn = 0,
-      scIndent = 0,
-      scThisLineSpans = [],
-      scRequestedDelimiter = case scRequestedDelimiter sc of
-        AfterNewline -> RequestedNewline
-        RequestedNewline -> RequestedNewline
-        VeryBeginning -> VeryBeginning
-        _ -> AfterNewline
-    }
+declNewline = newlineRawN =<< getPrinterOpt poNewlinesBetweenDecls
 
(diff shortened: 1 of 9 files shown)


wraithm:
@georgefst @parsonsmatt What do you think?

wraithm:
Also, all tests pass.

parsonsmatt:
I'm happy to support a style that you like!

src/Ormolu/Printer/Meat/Declaration.hs
```diff
@@ -68,12 +68,12 @@ p_hsDecls' grouping style decls =
...
       case grouping of
         Disregard ->
-          breakpoint : renderGroup curr
+          declBreakpoint : renderGroup curr
         Respect ->
           if separatedByBlankNE getLoc prev curr
             || isDocumented prev
             || isDocumented curr
-            then breakpoint : renderGroup curr
+            then declBreakpoint : renderGroup curr
```
georgefst, line 76: These `breakpoint`s could actually have been `newline` since we never print declarations on a single line. This means they can now be `declNewline`, and `declBreakpoint` needn't exist.

src/Ormolu/Printer/Internal.hs
```diff
@@ -365,7 +384,7 @@ newlineRaw = R . modify $ \sc ->
           scColumn = 0,
           scIndent = 0,
           scThisLineSpans = [],
-          scRequestedDelimiter = case scRequestedDelimiter sc of
+          scRequestedDelimiter = case requestedDel of
```
georgefst, line 387: I agree this is slightly cleaner, but since this function is otherwise untouched by this PR it should be reverted.

src/Ormolu/Printer/Internal.hs
```diff
@@ -316,6 +317,24 @@ space = R . modify $ \sc ->
...
+      scColumn = 0,
+      scIndent = 0,
+      scThisLineSpans = [],
+      scRequestedDelimiter = case scRequestedDelimiter sc of
+        AfterNewline -> RequestedNewline
+        RequestedNewline -> RequestedNewline
+        VeryBeginning -> VeryBeginning
+        _ -> AfterNewline
+    }
+
```
georgefst, line 337: There's a bit of repetition here with `newlineRaw`. Is there a good reason why this can't be written in such a way that `newlineRaw = newlineRawN 1`?

src/Ormolu/Printer/Meat/Module.hs
```diff
@@ -79,7 +79,7 @@ p_hsModule mstackHeader shebangs pragmas qualifiedPost HsModule {..} = do
     forM_ (normalizeImports preserveGroups hsmodImports) $ \importGroup -> do
       forM_ importGroup (located' (p_hsmodImport qualifiedPost))
       newline
-    newline
+    declNewline
```
georgefst, line 82: Curiously, the tests pass even with this line omitted entirely...

georgefst (changes requested):
This is great, and simpler than I expected. Just a few minor things.

src/Ormolu/Printer/Meat/Module.hs
georgefst: (I'm not suggesting actually changing anything here)

src/Ormolu/Printer/Internal.hs
wraithm: Of course! Yeah, I got overeager 👍

src/Ormolu/Printer/Meat/Declaration.hs
wraithm: 👍 np

src/Ormolu/Printer/Meat/Module.hs
wraithm: Redundant because two `newline` calls in a row is only one `newline`.

src/Ormolu/Printer/Internal.hs
wraithm: This is a great question. It all comes down to what these cases mean when there's no newline:

```
        { scBuilder = case requestedDel of
            AfterNewline -> builderSoFar
            RequestedNewline -> builderSoFar
            VeryBeginning -> builderSoFar
            _ -> builderSoFar <> "\n",
```

In my version of `newlineRawN`, I add newlines regardless of the value of `scRequestedDelimiter sc` is. In `newlineRaw`, `AfterNewline`, `RequestedNewline`, and `VeryBeginning` get nothing. Perhaps my version shouldn't do anything on `VeryBeginning` either, but I think because I'm explicitly trying to violate the invariant that `AfterNewline` and `RequestedNewline` newlines do nothing, it needs to be a separate function if we're going to maintain standard `ormolu` formatting in the body of functions or instance declarations or any other thing which might have multiple calls to `newlineRaw`, but want to not generate a newline.

I tried modifying `newlineRaw` to get the right behavior, but I couldn't really do it... Maybe the "correct" way to solve this problem is to have a new `scRequestedDelimiter` type which handles the end of declarations or something like that. However, I think that would be a pretty massive overhaul to the internals of ormolu that might not yield the right solution.

wraithm:
Okay @georgefst, I responded to all comments and pushed your requested changes 👍

src/Ormolu/Printer/Internal.hs
georgefst: I couldn't resist tinkering with this. See 071951f.

It's quite easy to see, via equational reasoning, that the new `newlineRaw` is equivalent to the old one.

And that commit doesn't change the behaviour with `poNewlinesBetweenDecls = 2` for anything in our, pretty extensive, test suite, so I'm fairly sure it's safe. But you may want to test with your own code.

wraithm: Okay! I'll do a test.

georgefst:
Ok @wraithm, I'll merge if you're happy with my new changes.

src/Ormolu/Printer/Internal.hs
wraithm: @georgefst This works great for me. I love it.

wraithm:
@georgefst I'm very happy. Go ahead!

wraithm:
🎉 Thanks!

wraithm:
@georgefst Would you mind making a hackage publish at some point soon? We're gonna convert to fourmolu at @bitnomial asap.

georgefst:
I suppose we could make a release with this and #44, which I'll hopefully get merged today.

I think PVP forces this to be a major release since it adds a record field (although I'd be surprised if Fourmolu were actually being used as a library anywhere other than in HLS, which I know wont be broken by this).

georgefst:
@wraithm [Done.](https://hackage.haskell.org/package/fourmolu-0.3.0.0/docs/Ormolu-Config.html#t:PrinterOpts)

wraithm:
You the best. Thanks!

PR #51, opened 2020-10-14
Remove hardcoded default for poNewlinesBetweenDecls from help text.

kukimik:
This just applies the machinery of #44 to the new option `--newlines-between-decls` introduced in #48 (so that we use the default value of `poNewlinesBetweenDecls` from `defaultPrinterOpts` instead of hardcoding it in the help text).


georgefst:
Good catch. I meant to check this, but I suppose I was lured in to a false sense of security by the lack of merge conflict.

wraithm:
Thanks!

PR #62, opened 2021-02-25
Relax aeson version lower bound requirement

gustavoavena:
As instructed in the README, ran `cabal test` and `./format.sh` and had no issues.

I would really appreciate if the hackage metadata could be updated after this is merged.

Thanks!

### DIFF
--- a/fourmolu.cabal
+++ b/fourmolu.cabal
@@ -117,7 +117,7 @@ library
 
     default-language: Haskell2010
     build-depends:
-        aeson >=1.5.2 && <1.6,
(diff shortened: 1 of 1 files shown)


georgefst:
Great, thanks. I've made a Hackage revision to match.

(The high lower bound is because we were at some point making use of `rejectUnknownFields`, but in the end that turned out to be annoying because it stopped config files from being backwards compatible, so was removed: 32f85d2. What we really want as haskell/aeson/808.)

gustavoavena:
Thanks @georgefst!

PR #64, opened 2021-03-04
Fix multiline haddocks in indented contexts

brandon-leapyear:
Fixes https://github.com/parsonsmatt/fourmolu/issues/63 
Fixes https://github.com/parsonsmatt/fourmolu/issues/53

### DIFF
--- a/src/Ormolu/Printer/Meat/Common.hs
+++ b/src/Ormolu/Printer/Meat/Common.hs
@@ -15,7 +15,6 @@ module Ormolu.Printer.Meat.Common
 where
 
 import Control.Monad
-import Data.Functor
 import Data.List (intersperse, isPrefixOf)
 import qualified Data.Text as T
 import GHC hiding (GhcPs, IE)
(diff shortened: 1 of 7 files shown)


src/Ormolu/Printer/Meat/Common.hs
```diff
@@ -171,11 +172,13 @@ p_hsDocString hstyle needsNewline (L l str) = do
           Asterisk n -> " " <> T.replicate n "*"
           Named name -> " $" <> T.pack name
         sequence_ $ intersperse (newline >> s) $ map txt' docLines
+
```
georgefst, line 175: I agree this is nicer with a few more line breaks. But I'm very weary of doing anything which unnecessarily increases the diff with upstream. Merging a new Ormolu version can be quite hairy already.

georgefst (changes requested):
This looks great. Thanks!

Just the minor quibble on whitespace.

src/Ormolu/Printer/Meat/Common.hs
brandon-leapyear: oof. completely understand. will revert

brandon-leapyear:
@georgefst is this fix worth a patch release to hackage?

georgefst:
We should hopefully be able to batch this up with a few other upcoming changes. Unless you could really do with it on Hackage for any reason?

brandon-leapyear:
If there are upcoming changes, then it can wait. As long as there's some release within a few weeks?

gustavoavena:
Is there an ETA for when this fix will be released?
Thanks!

brandon-leapyear:
@georgefst any updates on a Hackage release?

PR #90, opened 2021-05-20
do not revert Fourmolu magic comments to Ormolu form

### LINKED ISSUE #70: Disable pragmas always revert to Ormolu form
Formatting:
```hs
{- FOURMOLU_DISABLE -}
```

results in the output:
```hs
{- ORMOLU_DISABLE -}
```

As pointed out by @brandon-leapyear in #69.

I should really have thought of this when I added support (a15b229).

One option would just be to make `FOURMOLU` the [canonical version](https://github.com/parsonsmatt/fourmolu/blob/53a57a3bd7bec4beabff4612276365f07edcdd97/src/Ormolu/Processing/Preprocess.hs#L122). But in a sense this just moves the problem. Keeping track of which the user initially used could be mildly tricky.

kukimik:
Closes #70. 

After looking at the code I've decided **not** to add another option, and **not** to track which of the comments was used to disable the formatter. I just treat the two styles of comments as equivalent (e.g. after using `{- ORMOLU_DISABLE -}` we may use `{- FOURMOLU_ENABLE -}` to re-enable formatting), but I do not change the style of the comment used in the original file.

Adding an extra option just for this seemed weird. It's easy, but it adds much noise.
Tracking could be useful in some scenarios, but do we really need this?

This solution is simple, keeps Ormolu compatibility and is consistent with the current behavior.

However, if you think it's wrong this way, feel free to close the PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,7 +93,8 @@ source code must still be parseable even when the disabled regions are
 omitted. Because of that the magic comments cannot be placed arbitrarily,
 but rather must enclose independent top-level definitions.
 
-`{- ORMOLU_DISABLE -}` and `{- ORMOLU_ENABLE -}`, respectively, can be used to the same effect.
+`{- ORMOLU_DISABLE -}` and `{- ORMOLU_ENABLE -}`, respectively, can be used to the same effect,
+and the two styles of magic comments can be mixed.
 
 ## Current limitations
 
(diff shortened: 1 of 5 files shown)


georgefst:
Yeah, this is actually a better idea!

PR #101, opened 2021-06-25
Make the 'Loaded config from' message debugging output

rebeccaskinner:
What: 
Changes the "Loaded config from" message to only be printed alongside other debugging output

How:
Use `printDebug` instead of `hPutStrLn stderr` to print the message

Why:
1. Having clean output in the happy path makes editor integration somewhat easier. For example, using `shell-command-on-region` in emacs to format unsaved buffers causes emacs to insert the message from stderr into the file.
2. This is the only informational message in the application that is printed to `stderr` without debugging enabled. This change makes the behavior consistent with other message output.

### DIFF
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -369,7 +369,7 @@ mkConfig path Opts {..} = do
   filePrinterOpts <-
     loadConfigFile path >>= \case
       ConfigLoaded f po -> do
-        hPutStrLn stderr $ "Loaded config from: " <> f
+        printDebug $ "Loaded config from: " <> f
         printDebug $ show po
         return $ Just po
       ConfigParseError f (_pos, err) -> do


georgefst:
Given that this has also been requested in #110, I'm (somewhat reluctantly) happy to merge.

But then again, I don't really have a dog in this fight, as I pretty much only use Fourmolu via HLS, where we log everything, including debugging output, anyway. @brandonchinn178 thoughts?

brandonchinn178:
I also don't have a dog in the fight, but my preference would be to add a `--quiet` flag. I think it's nice to show what config is being loaded by default (in most cases, it's noise, but it some cases, it could prevent banging your head on the keyboard for an hour), but the fact that it's not configurable is an issue.

kenranunderscore:
I didn't know about the behavior with `shell-command-in-region`, but I created #110. I've also come to the conclusion that this kind of output is not a strange thing to do and that I probably should switch to a "real" formatter package (see https://github.com/haskell/haskell-mode/issues/1765).

Nevertheless, given the above-mentioned inconsistency, I'd love for any of those proposed changes to happen. I'd honestly be very happy with either solution (debug output vs. special flag) :) [The path of least resistance was using a patched `fourmolu` for now in a couple of projects.]

brandonchinn178:
Superceded by #170

Issue #71, opened 2021-03-10
Line length limit in fourmolu


gustavoavena:
**TL;DR:** Would an option for max number of characters in a line be a welcome feature in fourmolu? Is anyone asking about or working on this?

Based on ormolu's design principle's and a couple of comments that I've seen in its repo (e.g. [this](https://github.com/tweag/ormolu/issues/508)), I know that setting a limit to the number of character in a line is not something that they would want to do. 

FWIU, fourmolu was forked to provide more customization, so it now provides multiple options which I imagine wouldn't be considered in ormolu. With this in mind, I wanted to ask the maintainers about possibly adding an option for max number of characters in a line. 

- Would this be a welcome feature? Or is there a reason not to do this (e.g. increases the separation from ormolu, harder to maintain)?
- Has anyone asked for this before? Is anyone working or planning to work on this already?

Thanks!

georgefst:
> Would this be a welcome feature?

Pretty much any feature is welcome here! Unless it would be a real pain to maintain for any reason.

> Has anyone asked for this before?

Kind of: #19 

> Is anyone working or planning to work on this already?

No. I would like to see it, but it's fairly low on my list in terms of priority.

georgefst:
I had a hunch that this might be a more invasive change than anything else we've added, but it's hard to know without just trying to implement it.

parsonsmatt:
Sure! 

I'll even go against my own preferences here and allow people to customize it. As much as I might prefer the 80 column limit, I know other people prefer more (and sometimes less) ;)

gustavoavena:
Thank you both for the quick answers! I'm happy to hear that features like these would be welcome! 🙂

Yeah, I think that allowing users to set a max length would be a significant change... I'm not sure if I'll be able to work on something this big, but there might be some small tweaks that could achieve similar results (i.e. getting rid of very long lines and making the code easier to read).

FWIU, fourmolu is already able to break lines properly in things like lists or explicit import lists. The "issue" is that the user needs to break one of these lines for fourmolu to break and align the rest.

Maybe providing an option to enable automatic line-breaking for these cases (e.g. explicit imports, lists) could be a good start that would cover most of the long lines... WDYT?

My initial idea was to break lists with multiple elements if this option is enabled. Something like:

### Before
```haskell
import Foo (foo)
import Bar (bar1, bar2, bar3)

singleElementList :: [String]
singleElementList = ["one string"]

multipleElementsList :: [String]
[...]

PR #102, opened 2021-07-05
[RFC] Automatic line breaking at specific limit

gustavoavena:
Discussed in #71. @georgefst this is the "hack" you suggested, so I would love to get your feedback if you have some time!

### What
Provide an option to break lines automatically when they're over N characters long, where N is an optional integer provided in the config.

### How
Fourmolu already has logic for how to break lines, since it does so when the user breaks some lines themselves. This PR adds a small hack in `src/Ormolu/Printer/Combinators.hs` to use the MultiLine Layout for a span if it's over N characters long.

## Testing
Will add a file with some test cases in the next commit, but before putting up this PR I ran `cabal test` and `./format.sh`, as instructed in the `Contributing` section. The `./format.sh` script made some changes to the `fourmolu.cabal` file.

# RFCs / TODOs
- I struggled a bit to set up the new argument properly... I was able to make it work in the CLI (i.e. I tested both with and without the option in the CLI and it worked as expected), but the option is not being used when set **in a YAML file**. I would appreciate some help on how to add this option properly (i.e. it works both on the CLI and YAML config file) considering it should be a `Maybe Int`.
- Most of the test cases with this option set to 80 will have the same outcome as using the default fourmolu options. Instead of adding many duplicate files, I ran the tests using the `-four-out.hs` files and saw that only 3 failed because they have lines that are longer than 80 characters. So I generated the proper output files with the expected line breaking and modified the test runner (i.e. `PrinterSpec.hs`) to use the `-four-out.hs` files for all test cases, except those 3. I would love to get some feedback on that.
- I'm planning on adding a few more test cases with longer lines in a 3rd commit, e.g. long lists, long imports. Anything in particular that you think **needs** to be tested?


src/Ormolu/Config.hs
```diff
@@ -120,7 +120,9 @@ data PrinterOpts f = PrinterOpts
     -- | How to print doc comments
     poHaddockStyle :: f HaddockPrintStyle,
     -- | Number of newlines between top-level decls
-    poNewlinesBetweenDecls :: f Int
+    poNewlinesBetweenDecls :: f Int,
+    -- | Max line length for automatic line breaking
+    poMaxLineLength :: f (Maybe Int)
```
pepeiborra, line 125: I would create a new datatype for clarity: 

```
data ColumnLimit = ColumnLimit Natural | NoLimit
```

gustavoavena: Good idea. Just did that and pushed the changes. Yaml config file is still not working though, so I'll look into that.

gustavoavena:
Friendly ping for @georgefst. Would you mind doing an initial review when you have some time? 🙂

gustavoavena:
Fixed the YAML config file parsing for my new option. AFAIK, the only remaining work (besides comments from review) is to add another test case in a 3rd commit.

gustavoavena:
When writing some test cases I ended up finding a bug in the implementation. Apparently, it can't properly break function arguments when they're over 80 characters long.

Here are some examples of what works and what doesn't work (i.e. I get an error because the AST changed after formatting).

```haskell
-- ------------------------ These work ------------------------
-- ---------------------------------------------------------------80 characters|

longFunction0 :: String -> String -> String -> Maybe Int -> Maybe Int -> Maybe Int -> String -> [String]
longFunction0 veryLongArg1 a b c d e f = ["a list", "of strings", "that will break the", "column limit"]

longFunction1 :: String -> String -> String -> Maybe Int -> Maybe Int -> Maybe Int -> String
longFunction1 
    veryLongArg1 veryLongArg2 veryLongArg3 veryLongArg4 veryLongArgument5 veryLongArg6 = undefined

longFunction12 veryLongArg1 veryLongArg2 veryLongArg3 veryLongArg4 veryLongArgument5 
longFunction12 :: String -> String -> String -> Maybe Int -> Maybe Int -> Maybe Int -> String -> String
    veryLongArg6 veryLongArg6 = undefined


-- If the arguments stop right at 80 characters, it works.
longFunction2 :: String -> String -> String -> Maybe Int -> Maybe Int -> Maybe Int -> String
longFunction2 veryLongArg1 veryLongArg2 veryLongArg3 veryLongArg4 a veryLongArg6 =
    let aVeryLongLine = ["list one", "list one", "list one", "list one", "list one", "list one"]
    in undefined

-- ------------------------ Below here doesn't work ------------------------

longFunction3 :: String -> String -> String -> Maybe Int -> Maybe Int -> Maybe Int -> String
longFunction3 veryLongArg1 veryLongArg2 veryLongArg3 veryLongArg4 veryLongArgument5 veryLongArg6 = undefined
```

I formatted this using the `-u` flag to see the output and the last example (the one that breaks) becomes this:

```haskell
longFunction3 ::
    String -> String -> String -> Maybe Int -> Maybe Int -> Maybe Int -> String
longFunction3
veryLongArg1
veryLongArg2
veryLongArg3
veryLongArg4
veryLongArgument5
veryLongArg6 = undefined
```

I think that if the user doesn't break the arguments themselves, fourmolu can't distinguish the function name from the arguments... My only idea to make this work would be to disable this check for this specific case (function arguments). That would probably involve adding an argument to `switchLayout`, which calls `spansLayout`, to not use the column limit. The problem is that I don't have enough context on the code to find out where these functions will be called when parsing function definitions... 

@expipiplus @parsonsmatt  Could you maybe help me with that? Any code pointers that I could go over to disable the column limit in the `switchLayout` calls for function arguments?

gustavoavena:
I was actually able to find the place that I think `switchLayout` is used for the arguments list in the function definition. I just put up [a 3rd commit](https://github.com/fourmolu/fourmolu/pull/102/commits/6c75dc7c4e7cb0530508a222fda150beaf737685) with a fix that solved the issue.

gustavoavena:
I actually found a bug in this implementation that breaks idempotence in some cases... Since idempotence is an important property for a formatter, I don't want to ship this without a fix...
It seems to be related to the indentation of blocks after `do` statements in long lines (see examples at the bottom).

@georgefst, in one of your comments in #6 you said that "idempotence is a rather fragile thing". Were there any other features/improvements in the past that broke idempotence? If so, how were they fixed? I would appreciate anything that could help me debug and fix this issue. 

Thanks!



Original code:
```haskell
testFund :: Maybe Int
testFund = 
  firstTest oneFunctionArgument abcdeabcdeabcdeabcdeabcdeabcdeabcdeabcdeabcdeabcdeabcde $ do
    result <- aRandomResult
```

After formatting with column limit set to 80:
```haskell
testFund :: Maybe Int
testFund =
    firstTest
        oneFunctionArgument
        abcdeabcdeabcdeabcdeabcdeabcdeabcdeabcdeabcdeabcdeabcde $ do
        result <- aRandomResult
```

After formatting the output again:
```haskell
testFund :: Maybe Int
testFund =
    firstTest
        oneFunctionArgument
        abcdeabcdeabcdeabcdeabcdeabcdeabcdeabcdeabcdeabcdeabcde
        $ do
            result <- aRandomResult
```


Of course, formatting this snippet with the `-c` flag to check idempotence fails.

gustavoavena:
I tested fourmolu with this option in thousands of files and found out that this isn't a big idempotence issue because it has a stopping point. i.e. the formatter won't undo something it did and get into a vicious cycle of doing/undoing (e.g. adding/removing a line break). 

Also, this feature doesn't introduce any major issue that breaks the files (e.g. by changing their AST).

Considering what I've listed above and the fact that idempotence seems to be a very fragile property (IIRC, initial versions of Ormolu were not idempotent and it was still released and used with this known limitation), I was wondering if we could consider releasing this feature as long as it's clear that it has the limitation of affecting idempotence.

@kukimik, @brandonchinn178, I've seen you engage in other issues/PRs so I'm wondering if you have any thoughts about this? Any feedback would be very welcome! 🙂

georgefst:
I'm alright with the idempotence issue seeing as:
- we're fairly sure it doesn't get in to a loop
- the option is off by default
- the documentation of `--column-limit` makes the issue clear (maybe it'd be nice to categorise the kind of inputs that cause issues, or give examples?)

brandonchinn178:
Are you saying its not idempotent? As in, running fourmolu multiple times could iteratively format line breaks? Can you give an example?

In general, though, I'm generally in favor of this. Specifically with regards to breaking import statements.

gustavoavena:
Thank you @georgefst @brandonchinn178 for your responses and feedback!

@georgefst, I'm happy to hear that! I just pushed one more commit where I add an example in the test files, update the test runner and add a line to the README making the `column-limit` limitation clearer. WDYT?

@brandonchinn178 yes, in some cases the formatter will not be idempotent. I just added an example [here](https://github.com/fourmolu/fourmolu/pull/102/commits/fbeb895d475ad574b300eccfef7289321c154ae8). In practice though, it doesn't look like the file will change when formatted for a 3rd time and the changes between the first and second formatting are very minor (e.g. indentation of statements nested inside a `do` or a `$`).

data/examples/other/column-limit-tests-four-out.hs
```diff
@@ -63,3 +63,34 @@ longFunction2 veryLongArg1 veryLongArg2 veryLongArg3 veryLongArg4 a veryLongArg6
...
+-- Will become this when formatted with column-limit set to 80.
+testFund :: Maybe Int
+testFund =
+    firstTest
+        oneFunctionArgument
+        abcdeabcdeabcdeabcdeabcdeabcdeabcdeabcdeabcdeabcdeabcde
+        $ do
+            result <- aRandomResult
+
+-- Which, if formatted again, will become this:
```
brandonchinn178, line 89: this is the same as the above, right? mention this in the comment?

gustavoavena:
@georgefst LMK if there's anything else I need to add/change to merge this! 
I'm very excited to ship this, so people that hate long lines (like me) don't have to worry about breaking them manually anymore 😬

pepeiborra:
Is this ready to merge @brandonchinn178 @georgefst ?

gustavoavena:
Hey @georgefst! I noticed that there has been some significant changes committed in the past week, which is great!

I already rebased this PR and tested everything, so I'm wondering if we could get this merged?

ProofOfKeags:
Hi. What is needed for getting this merged? I want it :)

brandonchinn178:
👋 @gustavoavena 
Thank you for opening this PR and being patient while we sorted out how to handle all the new contributions for new configuration options!

If you haven't noticed, we finally merged in a new testing framework for configuration options ([#169](https://github.com/fourmolu/fourmolu/pull/169))! This means we can finally address the backlog of open PRs.

Please do the following steps

- [ ] Rebase against `main`

    ```bash
    git remote add upstream git@github.com:fourmolu/fourmolu
    git fetch upstream
    git rebase -i upstream/main
    ```

- [ ] Resolve merge conflicts
- [ ] Ensure `CHANGELOG.md` has been updated
- [ ] Ensure configuration docs in `README.md` have been updated
- [ ] Ensure `fourmolu.yaml` updated to stay in sync with config in `README.md`
- [ ] Ensure tests have been added

You can use [#135](https://github.com/fourmolu/fourmolu/pull/135) as a reference for this checklist.

gustavoavena:
@brandonchinn178 Thank you for the update and for all the you've been putting in recently. Unfortunately I haven't had the bandwidth to get back to this, but once I have some more time, I'll work on rebasing and writing test to ship this :)

arybczak:
Gentle ping - will work on getting this PR merged resume?

brandonchinn178:
@arybczak no ones currently working on this PR. feel free to branch off and start a new PR

CrystalSplitter:
I took up on the offer and branched off on this work in the attempt to revive the feature in #305. Thank you for your work, @gustavoavena!

georgefst:
This work has been merged via #305!

PR #106, opened 2021-08-10
Merge ormolu changes

brandonchinn178:
This merges in ormolu `master` as of today. This gets GHC 9 support, in addition to many other changes (this merges in 4 ormolu releases!)

Also adds `DEVELOPER.md` as a starting document for various developer workflows. Currently it has docs for the release process (feel free to edit as needed; i'm not sure how fourmolu was released in the past) and merging upstream ormolu changes

# After-merge checklist

- [ ] Add issue to unskip test: https://github.com/fourmolu/fourmolu/pull/106#discussion_r685807682
- [ ] Add issue to re-add regression test when ormolu fixes:
    - https://github.com/fourmolu/fourmolu/pull/106/commits/91470f23ac3b1ad49087ee172ff10225d492d485
    - https://github.com/tweag/ormolu/issues/474

Issue #49, opened 2020-10-13
Testing all options


georgefst:
Until recently, all our output options have been binary choices*, with a neat split in to "the Ormolu version" and "the Fourmolu version". Our tests have thus tested our default options, and Ormolu's default options. This was never going to last. #48 is the first to introduce an option which is not used by default.

We have a few choices:
- Run all the tests with all the combinations of options. This may take a long time to run (I haven't tried it yet), and it would start to become difficult to review all the outputs to check that everything looks sensible. But it does give us pretty comprehensive regression tests.
  - Obviously this grows exponentially as we add more options...
- Only test each option on the examples where it should be relevant, e.g. there's no point running with and without `record-brace-space` on a file that doesn't contain any records. Setting this up would require reviewing each test case to determine which options are relevant. We could also miss bugs where an option affects something it shouldn't.
- Consider which options are likely to interact with each other, and filter the set of combinations accordingly. This has some of the same drawbacks as the previous choice.

\* Apart from indentation, but for the purposes of this discussion that's a choice between two and four. There doesn't seem much point testing with any more than that.

kbertalan:
@georgefst The current tests are excellent for testing each option separately and isolated. I agree that checking those files with multiple with all possible combinations of options would be hard to maintain.

I think it would also be beneficial to provide few additional test cases, with more complex haskell files with ADTs, records, multiple top level declarations, imports, etc... Setting up those test files would be easy, just use few from the codebase, and maybe simplify if needed.

On these new test files (almost) all option combinations can be checked, if indentation and new lines between declarations have limited set of settings (2 or max 3).

As the number of output files count will be much higher, I advise to encode the options into the file name. An `example.hs` would have `example-i2-cst-iw-rbs-hssl-nbd1.hs` for the ormolu settings, which covers these settings:
* i2 - indent 2
* cst - comma style trailing
* iw - indent wheres enabled
* rbs - record brace space enabled
* diff friendly import export disabled (if enabled it would be `dfie`)
* respectful disabled (if enabled it would be `r`)
* hssl - haddock style single line
* nbd1 - new lines between declarations set to 1

With this approach there would be at least 256 outputs for a single test file, so skipping some combinations still would be needed to reduce the maintenance cost.

Can you comment on this idea?

kbertalan:
I have created a fork and experimented with the approach I described above. It is located at https://github.com/kbertalan/fourmolu/tree/testing-all-options

georgefst:
[...]

PR #108, opened 2021-08-24
Add leading commas option for module import-export lists

### LINKED ISSUE #61: `comma-style: leading` doesn't apply to import and export lists
I get:
```haskell
module Logging (
  LogAction,
  logMessageStderr,

  -- * Messages
  Message (..),
  HasLog (..),
  WithLog,
  logDebug,
  logInfo,
  logWarning,
  logError,

  -- * Message Severity
  module Colog.Core.Severity,
) where
```

I want:
```haskell
module Logging
  ( LogAction
  , logMessageStderr

  -- * Messages
  , Message(..)
  , HasLog(..)
  , WithLog
  , logDebug
  , logInfo
  , logWarning
  , logError

  -- * Message Severity
  , module Colog.Core.Severity
  ) where
```

This seems like a bug in the option, personally.

3kyro:
The implementation passes all tests when import-export comma style is set to trailing, and only fails expected formatting tests when set to leading.

Probably needs to wait #106 to be merged first.

Fixes #61


3kyro:
Formatting when comma style is set to leading:

````haskell
       >>>>>>>>>>>>>>>>>>>>>> trailing:
       module Foo
         ( foo,
           bar,
           baz,
         )
       where

       >>>>>>>>>>>>>>>>>>>>>> leading:
       module Foo
         ( foo
         , bar
         , baz
         )
       where
````

````haskell
       >>>>>>>>>>>>>>>>>>>>>> trailing:
       import qualified MegaModule as M
         ( Either,
           Maybe (Just, Nothing),
           MaybeT (..),
           Monad (return, (>>), (>>=)),
           MonadBaseControl,
           join,
           liftIO,
           void,
           (<<<),
           (>>>),
         )

       >>>>>>>>>>>>>>>>>>>>>> leading:
       import qualified MegaModule as M
         ( Either
         , Maybe (Just, Nothing)
         , MaybeT (..)
         , Monad (return, (>>), (>>=))
         , MonadBaseControl
         , join
         , liftIO
         , void
         , (<<<)
         , (>>>)
         )
````


````haskell
>>>>>>>>>>>>>>>>>>>>>> trailing:
       {-# LANGUAGE FlexibleContexts #-}
       {-# LANGUAGE TypeFamilies #-}

       -- | Header.
       module My.Module
         ( -- * Something
           foo,
           bar,
           -- | A multiline
           -- comment here
           bar2,

           -- * Another thing
           (<?>),
           {- some other thing -} foo2, -- yet another
           foo3, -- third one
           baz,
           bar2, -- a multiline comment
           -- the second line
           bar3,
           module Foo.Bar.Baz,
         )
       where

       -- Wow

       >>>>>>>>>>>>>>>>>>>>>> leading:
       {-# LANGUAGE FlexibleContexts #-}
       {-# LANGUAGE TypeFamilies #-}

       -- | Header.
       module My.Module
         ( -- * Something
         foo
         , bar
         -- | A multiline
         -- comment here
         , bar2

         -- * Another thing
         , (<?>)
         {- some other thing -} , foo2 -- yet another
         , foo3 -- third one
         , baz
         , bar2 -- a multiline comment
         -- the second line
         , bar3
         , module Foo.Bar.Baz
         )
       where

       -- Wow
````

**Comment by 3kyro** [comment]
I fumbled a bit with the configuration options, but should be good now. 😄

**Comment by novadev94** [comment]
Hi, what forbids this PR from getting merged? I'm glad to help.

**Comment by 3kyro** [comment]
> Hi, what forbids this PR from getting merged? I'm glad to help.

It kind of got stuck waiting for #49 . We are actually considering whether to use fourmolu at work, and having this option is important. @georgefst  Is there something else that would forbid this to get merged?

File: src/Ormolu/Printer/Meat/ImportExport.hs [hunk 1]
```diff
@@ -75,41 +76,40 @@ p_hsmodImport ImportDecl {..} = do
...
-p_lie encLayout relativePos = \case
-  IEVar NoExtField l1 -> do
-    located l1 p_ieWrappedName
-    p_comma
-  IEThingAbs NoExtField l1 -> do
-    located l1 p_ieWrappedName
-    p_comma
-  IEThingAll NoExtField l1 -> do
+p_lie :: Layout -> RelativePos -> CommaStyle -> IE GhcPs -> R ()
+p_lie encLayout relativePos commaStyle = \case
```
brandonchinn178, line 87: To avoid conflicts with upstream, can we avoid adding a new parameter here, and instead get the comma style in `withComma`? It should be as simple as
```diff
 MultiLine -> do
-  case commaStyle of
+  getPrinterOpt poIECommaStyle >>= \case
     Leading -> ...
```

src/Ormolu/Printer/Meat/ImportExport.hs
```diff
@@ -75,41 +76,40 @@ p_hsmodImport ImportDecl {..} = do
...
-      parens' False . sep commaDel' sitcc $
-        case w of
-          NoIEWildcard -> names
-          IEWildcard n ->
-            let (before, after) = splitAt n names
-             in before ++ [txt ".."] ++ after
-    p_comma
-  IEModuleContents NoExtField l1 -> do
+  IEThingWith NoExtField l1 w xs _ -> sitcc $
+    withComma $ do
```
brandonchinn178, line 99: To avoid conflicts with upstream, can this just be
```hs
sitcc . withComma $ do
```
?

fourmolu.yaml
```diff
@@ -1,5 +1,6 @@
 indentation: 2
 comma-style: trailing
+ie-comma-style: trailing
```
brandonchinn178, line 3: Personally, I find this config name to be a bit too esoteric. Can we just be verbose and do `import-export-comma-style`? (I would also likewise rename `poIECommaStyle`, but that's less important because it's developer-facing only)

data/examples/module-header/multiline-with-comments-four-ie-out.hs
```diff
@@ -0,0 +1,25 @@
+{-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE TypeFamilies #-}
+
+-- | Header.
+module My.Module
+    ( -- * Something
+    foo
```
brandonchinn178, line 7: This is the one functional concern I have with this PR: I would personally expect/prefer `foo` to be at the same position as `bar`. Is that possible currently?

Possibly related: https://github.com/fourmolu/fourmolu/issues/145

brandonchinn178 (commented):
Reference for the comments about conflicting with upstream: https://github.com/fourmolu/fourmolu/blob/master/DEVELOPER.md#contributing

Renaming the config option is my biggest concern otherwise. I'll leave the new tests (and decision on whether to block on #49) to @georgefst

src/Ormolu/Printer/Meat/ImportExport.hs
3kyro: Thanks, indeed it's much simpler. done.

src/Ormolu/Printer/Meat/ImportExport.hs
3kyro: thanks, done

fourmolu.yaml
3kyro: I agree, it's not a very good choice for an option name. Changed as per your suggestion

data/examples/module-header/multiline-with-comments-four-ie-out.hs
3kyro: I've made some modifications to explicitly indent an export item after documentation. I've also moved all comments and docs to the same column, leaving only commas at the parentheses column. Only multiple line comments that don't start on a newline are left as were, since changing them would require some internal changes that I don't believe are worth the trouble. My comment above with examples has been updated as well.

3kyro:
> Reference for the comments about conflicting with upstream: https://github.com/fourmolu/fourmolu/blob/master/DEVELOPER.md#contributing
> 
> Renaming the config option is my biggest concern otherwise. I'll leave the new tests (and decision on whether to block on #49) to @georgefst

Regarding testing, this PR adds new test files where necessary that test output only with the new option enabled. Regarding #49 I kind of feel that a leading commas option import export lists is more of a continuation of the way fourmolu fundamentally differs from ormolu regarding lists rather than a specific one way option. That being said, I'm happy to conform to a specific test format when one is decided. @georgefst I would really appreciate your input.

tchoutri:
@brandonchinn178 Anything I could do to help this PR getting merged? :) I'm also oscillating towards Fourmolou thanks to this PR.

brandonchinn178:
All of my comments have been addressed. It's up to @georgefst on what to do about testing

tchoutri:
Thank you very much. :)

parsonsmatt (approved):
I’m in favor and would love to have this feature in.

I’m sympathetic to the testing argument. Obviously testing the combinatorics possibilities of all configuration options isn’t tenable. But I think it should be possible to trim that state space considerably

arybczak:
Ping. Could this be merged?

brandonchinn178:
@georgefst Would you be ok merging this now? Also, can you email me so we can briefly chat about how to coordinate changes to fourmolu? (my emails in my GitHub profile)

cc @parsonsmatt if you want to be included in that conversation as well, email me and I'll keep you in the loop

georgefst:
> @georgefst Would you be ok merging this now?

Sure. I'm willing to make an exception in this case, due to the PR containing its own tests, and the sheer level of demand as evidenced by comments and reactions. Besides, code changes (other than tests) are fairly small and @brandonchinn178 appears to have reviewed thoroughly.

> Also, can you email me so we can briefly chat about how to coordinate changes to fourmolu?

Done.

georgefst:
> I kind of feel that a leading commas option import export lists is more of a continuation of the way fourmolu fundamentally differs from ormolu regarding lists rather than a specific one way option

There **is** an important difference here between import/export lists, and list literals. Which is that the former allows extra commas, which means we can make the trailing style more uniform.

In fact, I'd argue that they're two quite separate concepts that just happen to have similar syntax, and thus there isn't any great argument against using trailing commas for one and leading commas for the other. So I feel quite strongly that the two options should be separate, and that we should maintain the current default behaviour (which, of course, is what this PR now does).

(I've made this point somewhere before in a related conversation, but I can't find it right now)

tchoutri:
@georgefst Thank you for this important precision, this is indeed something that I wasn't quite aware of. :)

jhrcek:
I'm very happy this feature is now available.
Are there any plans to release new version of fourmolu with this feature?

brandonchinn178:
I'm planning on doing a release in a couple weeks, is that ok?

brandonchinn178:
FYI `import-export-comma-style=Leading` looks a bit weird with `diff-friendly-import-export=True`:
```hs
module Foo (
    foo
    , bar
    , baz
) where

import qualified MegaModule as M (
    Either
    , Maybe (Just, Nothing)
    , MaybeT (..)
    , Monad (return, (>>), (>>=))
    , MonadBaseControl
    , join
    , liftIO
    , void
    , (<<<)
    , (>>>)
 )
```
Not sure how this option should behave in this case

tchoutri:
This seems appropriate:

```haskell
module Foo (
      foo
    , bar
    , baz
) where
```

georgefst:
> This seems appropriate:
> 
> ```haskell
> module Foo (
>       foo
>     , bar
>     , baz
> ) where
> ```

That is probably the best we can do, which neatly indicates that the "diff-friendly"-ness of the latter goes out the window with leading commas. So using both together seems undesirable.

Hmm. Ideally these would be combined in to a single ternary option, but maybe it's not worth the breakage. Otherwise we can just mention in the option docs (which I think are being worked on in #169) that they interact weirdly.

tchoutri:
@georgefst Yes, I think being truthful about the interactions might be the best way to continue. :)

brandonchinn178:
Something like `import-export-style: diff-friendly, leading, trailing` would be nice. I'll keep it in mind if we ever add more options to this area

Issue #114, opened 2021-09-30
Fix test for same-line do-block


brandonchinn178:
Uncomment this test:
https://github.com/fourmolu/fourmolu/blob/4f40026004a734fd5f2e6884391756bfcc9dc0c2/data/examples/declaration/value/function/application-2.hs#L11-L14

PR #115, opened 2021-09-30
Fix same line do block

brandonchinn178:
Resolves https://github.com/fourmolu/fourmolu/issues/114

### DIFF
--- a/src/Ormolu/Printer/Internal.hs
+++ b/src/Ormolu/Printer/Internal.hs
@@ -21,6 +21,7 @@ module Ormolu.Printer.Internal
     declNewline,
     useRecordDot,
     inci,
+    inciBy,
     inciByFrac,
     inciHalf,
     sitcc,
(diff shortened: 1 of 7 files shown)


src/Ormolu/Printer/Internal.hs
```diff
@@ -398,30 +399,35 @@ newlineRawN n = R . modify $ \sc ->
...
-          { rcIndent = roundDownToNearest step (rcIndent rc) + step
-          }
-  R (local modRC m)
+inciBy step (R m) = R (local modRC m)
   where
+    modRC rc =
+      rc
+        { rcIndent = roundDownToNearest step (rcIndent rc) + step
+        }
     roundDownToNearest r n = (n `div` r) * r
```
brandonchinn178, line 410: now this matches Ormolu's version (modulo `roundDownToNearest`)

src/Ormolu/Printer/Meat/Declaration/Value.hs
```diff
@@ -604,7 +604,9 @@ p_hsExpr' s = \case
             -- Thus, we indent by half of indentStep when the function is
             -- a multi line do block or case expression.
             indentArg
-              | isOneLineSpan (getLoc func) = inci
+              | isOneLineSpan (getLoc func) = case unLoc func of
+                  HsDo {} -> inciBy 2
+                  _ -> inci
```
brandonchinn178, line 609: This worked in Ormolu because their indentation size is hardcoded to 2. It seems like for the case of a do-block in a one-line span, the indentation size should be exactly 2 (or 1 might work), regardless of the indentation setting

georgefst (approved):
Nice!

Issue #111, opened 2021-09-20
Better documentation of formatting options


georgefst:
We ought to explain more clearly what each option does, with a side-by-side comparison of a minimal example. We could also explain why one might prefer one setting over another.

For now, this should probably exist as a markdown document in this repo, separate from the README.

georgefst:
Make sure to tackle #56.

georgefst:
@brandonchinn178 has made significant improvements to this documentation since this issue was opened. And #239 improves things further. But I still think it would be best to see nicely marked-up simple examples right there next to the descriptions. Trawling through the test suite is a lot more work.

brandonchinn178:
I believe the website has good docs for formatting options now 😄

Issue #33, opened 2020-09-29
Rename 'preserve-spacing'


georgefst:
Thinking ahead a little, I can see us wanting a more general "preserve more user choice" flag, to make Fourmolu less keen to rearrange comments etc.

Unsure about a good name.

PR #117, opened 2021-10-04
Miscellaneous small improvements, mostly documentation

georgefst:
All other than 329363f59de18cd75b6d6b6b46f4c5be511b6954 are documentation improvements.

### DIFF
--- a/DEVELOPER.md
+++ b/DEVELOPER.md
@@ -8,6 +8,36 @@ Some things to keep in mind when making changes:
     * Avoid refactoring where possible, don't reformat untouched code
     * Since we continuously merge in changes from Ormolu, reducing the number of potential conflicts goes a long way towards maintainability of this project.
 
+## Instant feedback with GHCID
+
+We often want to immediately see how changes to Fourmolu's source code affect outputs. Try adding something like this to `Ormolu.hs`:
+
+```hs
+import qualified Data.Text.IO as T
+import System.Directory (getHomeDirectory)
+import System.FilePath ((</>))
+
+main :: IO ()
+main = do
+  dir <- (</> "Desktop") <$> getHomeDirectory
+  ormoluFile conf (dir </> "In.hs") >>= T.writeFile (dir </> "Out.hs")
+  where
+    conf =
+      defaultConfig
+        { cfgUnsafe = True,
+          cfgPrinterOpts =
+            defaultPrinterOpts
+              { poCommaStyle = pure Trailing
+              }
+        }
+```
+
+Put some interesting code in `In.hs`. The contents of `Out.hs` can be kept up to date to reflect the result of running Fourmolu on it, by running:
+
+```
+ghcid -c 'cabal repl' -W -r --reload=$HOME/Desktop/In.hs
+```
+
 ## Release a new version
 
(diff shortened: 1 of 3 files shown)


fourmolu.yaml
```diff
@@ -3,5 +3,5 @@ comma-style: trailing
 record-brace-space: true
 indent-wheres: true
 diff-friendly-import-export: false
-respectful: false
+respectful: true
```
brandonchinn178, line 6: I'm not entirely sure what this option does. Does the fact that there are no formatting changes in this PR mean that `respectful` doesn't do any changes today, but we would like to start making breaking changes in `respectful=false`?

README.md
```diff
@@ -50,7 +50,18 @@ haddock-style: multi-line # '--' vs. '{-'
 newlines-between-decls: 1 # number of newlines between top-level declarations
 ```
 
-See [here](fourmolu.yaml) for a config to simulate the behaviour of Ormolu.
+A config to simulate the behaviour of Ormolu ([as far as currently possible](https://github.com/fourmolu/fourmolu/issues/38)) would be:
```
brandonchinn178, line 53: I don't understand this. Shouldn't we keep formatting fourmolu source code the same way Ormolu does, to avoid merge conflicts? If so, then the in-repo fourmolu config should be exactly the configuration needed to simulate Ormolu's formatting

DEVELOPER.md
```diff
@@ -104,3 +104,9 @@ Fourmolu aims to continue merging upstream changes in Ormolu. Whenever Ormolu ma
         fi
     done
     ```
+
+## HLint
+
+Ormolu isn't HLint-clean, so Fourmolu can't be fully.
```
brandonchinn178, line 140: :( Is Ormolu open to adding this?

brandonchinn178 (approved):
Other than some questions, looks good to go

DEVELOPER.md
georgefst: They weren't [at some point](https://github.com/tweag/ormolu/pull/455#discussion_r347565325). Maybe that conversation could be revisited, but I don't really have the energy for it.

brandonchinn178: For sure. No worries, just curious

README.md
georgefst: It's an expected property of the `--respectful` flag that it won't change existing code any more than Ormolu does (I should document this when tackling #111). It's possible that we might want to take advantage of Ormolu's increased flexibility in code we write ourselves. My motivation was bb3cbc1086cb241e6c833c6456874cab952afb56, where the import groups get merged without `--respectful`.

fourmolu.yaml
georgefst: > I'm not entirely sure what this option does.

It's a slight pun on the idea that the formatter should "respect" more user choice. e.g. https://github.com/tweag/ormolu/issues/74. There's some background in #33 but admittedly I've never made the behaviour totally clear. It's the main flag I had in mind when opening #111.

> Does the fact that there are no formatting changes in this PR mean that `respectful` doesn't do any changes today, but we would like to start making breaking changes in `respectful=false`?

No. See my other comment.

PR #123, opened 2021-10-15
Fix indentation on guard indentation

### LINKED ISSUE #122: Fix indentation with multiple guard conditions
Currently, fourmolu outputs:
```hs
-- indentation=2
foo mx my
  | Just x <- mx
    , Just y <- my =
    x + y

-- indentation=4
foo arr
    | Just x <- mx
      , Just y <- my =
        x + y
```
It should probably instead be
```diff
 -- indentation=2
 foo mx my
   | Just x <- mx
-    , Just y <- my =
+  , Just y <- my =
     x + y

  -- indentation=4
 foo arr
     | x <- arr
-       , y <- arr =
+    , y <- arr =
         x + y
```
As a bonus, it might be nice for `indentation=2` to indent the body one more, e.g.
```diff
 foo mx my
   | Just x <- mx
   , Just y <- my =
-    x + y
+      x + y
```
to visually separate the body from the guards even more, especially in something like
```hs
case mx of
  Just x
    | x > 10
    , x < 20 -> do
      print x
      return (x + 2)
    | otherwise -> do
      print (x + 2)
      return x
  Nothing -> return ()

-- vs

case mx of
  Just x
    | x > 10
    , x < 20 -> do
        print x
        return (x + 2)
    | otherwise -> do
        print (x + 2)
        return x
  Nothing -> x
```

ref: https://github.com/fourmolu/fourmolu/pull/121#discussion_r729298937

brandonchinn178:
Fixes #122

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## Unreleased
+
+* Fixed issue with leading commas in guards ([#123](https://github.com/fourmolu/fourmolu/pull/123))
+
 ## Fourmolu 0.4.0.0
 
 * Write "Loaded config from" message to `stderr` instead of `stdout` ([#42](https://github.com/fourmolu/fourmolu/pull/42))


CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 * Write "Loaded config from" message to `stderr` instead of `stdout` ([#42](https://github.com/fourmolu/fourmolu/pull/42))
 * Fixed issue with multiline haddocks in indented contexts ([#64](https://github.com/fourmolu/fourmolu/pull/64))
 * Avoid renaming magic comments ([#90](https://github.com/fourmolu/fourmolu/pull/90))
+* Fixed issue with leading commas in guards ([#123](https://github.com/fourmolu/fourmolu/pull/123))
```
georgefst, line 6: This should go in a new `Unreleased` section above. We've already released 0.4.0.0.

georgefst (changes requested):

CHANGELOG.md
brandonchinn178: ah yes sorry, fixed

PR #98, opened 2021-06-15
WIP: Add option for trailing haddocks in records

expipiplus1:
TODO: Tests

TODO: Disallow this in combination with trailing comma style

Issue #112, opened 2021-09-22
Merge pull requests


pepeiborra:
There are 12 open PRs right now, and at least 4 of them have a green light meaning all the tests pass and there are no merge conflicts. The oldest one is 3 months old and the other ones are all >1 month old. (#108, #103, #102, #101)

My concern is that the maintainers are no longer able to work on fourmolu. Maybe it would be a good idea to issue a call for new maintainers?

jneira:
Pinging @georgefst, as they have been merge last prs here

georgefst:
Thanks @jneira. I have seen this, and started writing up a full response on Wednesday, but haven't yet had the time to finish it. Which, actually, probably tells you most of what you need to know.

The tl;dr will be that I'm sufficiently motivated and available to just about keep the project ticking over, but we could really do with a few more pairs of hands.

georgefst:
Many open PRs are effectively blocked due to the lack of testing: see https://github.com/fourmolu/fourmolu/issues/49#issuecomment-955271097.

georgefst:
As many of you will have noticed, @brandonchinn178 has taken over much of the day-to-day maintenance of the project. I'm still keeping an eye on things (and to a lesser extent, so is @parsonsmatt), and commenting on opened issues and PRs where I feel I have useful input, but I expect I'll mostly continue to struggle to find the time to do any implementation work or detailed code reviews.

Anyway, thanks largely to Brandon's hard work, we're in much better shape than when this issue was opened. In particular, we have implemented a solution to the testing problem in https://github.com/fourmolu/fourmolu/pull/169. Some PRs which were blocked on that unfortunately remain unmerged, because their original authors haven't been able to get back to them yet. But if anyone here cares a lot about any particular PR, feel free to pick it up and carry on the work yourself. Brandon has even left a helpful comment on each one about what needs to be done to bring the work up to date and add tests.

# Some background

(This was going to form the basis of a mea culpa of sorts, alluded to in https://github.com/fourmolu/fourmolu/issues/112#issuecomment-927383812, but a lot of what I was going to say isn't relevant 10 months on.)

[...]

Issue #72, opened 2021-03-13
Comments that should not be moved between leading comma and element


brandon-leapyear:
We have comments that separate out elements in our list, like
```hs
[ foo1
  -- bars are below here
, bar1
, bar2
, bar3
  -- bazes below here
, baz1
, baz2
]
```
but Fourmolu reformats this to
```hs
[ foo1
, -- bars below here
  bar1
, bar2
, bar3
, -- bazes below here
  baz1
, baz2
]
```
which is weird because it seems like it's commenting that specific element rather than everything after that point. Ormolu doesn't have this problem because it doesn't have the leading commas
```hs
-- ormolu
[ foo1,
  -- bars below here
  bar1,
  bar2,
  -- bazes below here
  baz1,
  baz2
]
```

brandon-leapyear:
Also applicable for sum types
```hs
data Foo
  = Foo
  | Bar
  -- other constructors
  | Other1
  | Other 2
```

expipiplus1:
I bumped into something similar during my leading arrow work and made it so that comments were sensibly attached. https://github.com/fourmolu/fourmolu/blob/35dbface79d722952163e0c862951cbd6ca175fc/src/Ormolu/Printer/Meat/Type.hs#L144-L145

A workaround is to add a comment after `foo1` and `bar2` like so

```haskell
x =
  [ foo1 --
  -- bars are below here
  , bar1
  , bar2
  , bar3 --
  -- bazes below here
  , baz1
  , baz2
  ]
```

expipiplus1:
This fixes it I believe 

```haskell
diff --git a/src/Ormolu/Printer/Meat/Declaration/Value.hs b/src/Ormolu/Printer/Meat/Declaration/Value.hs
index 6443466..77e1597 100644
--- a/src/Ormolu/Printer/Meat/Declaration/Value.hs
+++ b/src/Ormolu/Printer/Meat/Declaration/Value.hs
@@ -720,7 +720,13 @@ p_hsExpr' s = \case
       TransStmtCtxt _ -> notImplemented "TransStmtCtxt"
   ExplicitList _ _ xs ->
     brackets s $
-      sep commaDel (sitcc . located' p_hsExprListItem) xs
+      getPrinterOpt poCommaStyle >>= \case
+        Leading ->
+          let ps = pure () : repeat (comma >> space)
+              xs' = zip ps xs
[...]

PR #124, opened 2021-10-30
Use trailing haddocks for record fields when using leading commas

georgefst:
This is inspired by (and lifts the core code from) #98.

Differences:
- It's rebased on to `master`.
- The comments are moved two spaces left. I think this looks better, but I'm happy to hear any dissenting opinions. This is the reason for the removal of the `sitcc` call in `Ormolu.Printer.Meat.Type`.
- There's no separate option. Instead we just follow `poCommaStyle`, and use trailing haddocks iff it's set to `Leading`. This removes the "Disallow this in combination with trailing comma style" concern (I'm not sure `optparse-applicative` provides a nice way to do that). And the old style is so ugly that I really don't see any reason why anyone would prefer it. It's just a relic of adapting Ormolu to leading commas in the simplest way possible.

### DIFF
--- a/data/examples/declaration/data/record-four-out.hs
+++ b/data/examples/declaration/data/record-four-out.hs
@@ -2,22 +2,22 @@ module Main where
 
 -- | Something.
 data Foo = Foo
-    { -- | X
-      fooX :: Int
-    , -- | Y
-      fooY :: Int
-    , -- | BarBaz
-      fooBar, fooBaz :: NonEmpty (Identity Bool)
-    , -- | GagGog
-      fooGag
+    { fooX :: Int
+    -- ^ X
+    , fooY :: Int
+    -- ^ Y
+    , fooBar, fooBaz :: NonEmpty (Identity Bool)
+    -- ^ BarBaz
+    , fooGag
       , fooGog ::
         NonEmpty
             ( Indentity
                 Bool
             )
-    , -- | Huh!
-      fooFoo
+    -- ^ GagGog
+    , fooFoo
       , barBar ::
         Int
+    -- ^ Huh!
     }
     deriving (Eq, Show)
--- a/data/examples/declaration/data/field-layout/record-four-out.hs
+++ b/data/examples/declaration/data/field-layout/record-four-out.hs
@@ -2,11 +2,11 @@ module Main where
 
 -- | Foo.
 data Foo = Foo
-    { -- | Something
-      foo :: Foo Int Int
-    , -- | Something else
-      bar ::
+    { foo :: Foo Int Int
+    -- ^ Something
+    , bar ::
         Bar
             Char
             Char
+    -- ^ Something else
     }
--- a/data/examples/declaration/data/record-multi-const-four-out.hs
+++ b/data/examples/declaration/data/record-multi-const-four-out.hs
@@ -3,15 +3,15 @@ module Main where
 -- | Something.
 data Foo
     = Foo
-        { -- | X
-          fooX :: Int
-        , -- | Y
-          fooY :: Int
+        { fooX :: Int
+        -- ^ X
+        , fooY :: Int
+        -- ^ Y
         }
     | Bar
-        { -- | X
-          barX :: Int
-        , -- | Y
-          barY :: Int
+        { barX :: Int
+        -- ^ X
+        , barY :: Int
+        -- ^ Y
         }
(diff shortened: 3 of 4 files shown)


expipiplus1:
>     * The comments are moved two spaces left. I think this looks better, but I'm happy to hear any dissenting opinions. This is the reason for the removal of the `sitcc` call in `Ormolu.Printer.Meat.Type`.

I actually quite prefer to have them indented from the commas, IMO having the commas 'sticking out' like that helps separate the members at glance. When they're lined up with the commas it's kind of one large block; distinguishing dashes from commas is harder than distinguishing space from commas.

I suppose the 'most fourmolu' thing to do would be to make this configurable.

image

georgefst:
> I suppose the 'most fourmolu' thing to do would be to make this configurable.

Ok, sure. Any opinion on naming the flag? I can't think of anything snappier than `--indent-trailing-haddocks`.

expipiplus1:
Seems fine to me, it's not as if anyone actually types config options out
for a formatter anyway (aside from putting them in their dotfiles once
ever).

Does make me think that my preference of indenting for records but not for
functions is a little inconsistent...

georgefst:
> Seems fine to me, it's not as if anyone actually types config options out for a formatter anyway (aside from putting them in their dotfiles once ever).

Good point.

> Does make me think that my preference of indenting for records but not for functions is a little inconsistent...

Yeah, it looks pretty weird to me. And doing the analogous thing for functions would require a 3-space indent... But, hey, if you want the option, I'm happy to add it. That's basically our mission statement.

expipiplus1:
> Yeah, it looks pretty weird to me. And doing the analogous thing for functions would require a 3-space indent... But, hey, if you want the option, I'm happy to add it! That's basically our mission statement.

So as to not be difficult, I'll try this out your way, and if it's really as terrible as my knee-jerk reaction, then I'll implement an option for it myself.

data/examples/declaration/data/record-four-out.hs
```diff
@@ -2,22 +2,22 @@ module Main where
...
-    , -- | Y
-      fooY :: Int
-    , -- | BarBaz
-      fooBar, fooBaz :: NonEmpty (Identity Bool)
-    , -- | GagGog
-      fooGag
+    { fooX :: Int
+    -- ^ X
+    , fooY :: Int
+    -- ^ Y
```
brandon-leapyear, line 8: Personally, I would imagine it lining up better as
```hs
{ fooX :: Int
  -- ^ X
, fooY :: Int
  -- ^ Y
```
(i.e. same indentation + 2 spaces, regardless of indentation size)

More importantly, if a user did
```hs
{ fooX :: Int -- ^ X
, fooY :: Int -- ^ Y
```
it seems like fourmolu forces these on the next line. I would expect fourmolu to respect the user's newline here (or lack thereof)

georgefst: 1) Fair enough, if you and @expipiplus1 both prefer that I'd be happy to support it as an option if someone wants to implement it (once we start clearing the backlog anyway: https://github.com/fourmolu/fourmolu/issues/112#issuecomment-955272628).

2) That really isn't ideal - I'll look in to it.

brandonchinn178: 1. Ah I didn't notice it's what @expipiplus1 said. Yeah, no worries, I'm not _terribly_ bothered by it. I'll open a ticket
2. I'll open a ticket for tracking it

brandonchinn178: Oh this also causes weird behavior when trying to comment sections of record fields:

Original:
```hs
data Foo = Foo
  { -- section 1
    a :: Int
  , b :: Int
  , c :: Int
    -- section 2
  , d :: Int
  , e :: Int
  }
```
`fourmolu-0.4.0.0`:
```diff
data Foo = Foo
  { -- section 1
    a :: Int
  , b :: Int
  , c :: Int
-   -- section 2
- , d :: Int
+ , -- section 2
+   d :: Int
  , e :: Int
  }
```
`fourmolu-0.5.0.0`:
```diff
data Foo = Foo
  { -- section 1
    a :: Int
  , b :: Int
  , c :: Int
-   -- section 2
- , d :: Int
+ , -- section 2
+ d :: Int
  , e :: Int
  }
```

Related: #72

PR #125, opened 2021-10-30
Use trailing haddocks for constructors in datatype declarations

georgefst:
It is a bit unfortunate that this forces single-line layouts to multi-line when there's a haddock on the constructor. As far as I'm aware this sort of behaviour is unprecedented.

Perhaps the leading style like https://github.com/tweag/ormolu/issues/583#issuecomment-644043287 is the ideal. Although the fact that comments then appear before the `|` is arguably slightly confusing and may hamper readability. Also, as indicated in https://github.com/tweag/ormolu/issues/583#issuecomment-646550697 it means a significant refactor may be necessary. Perhaps if one of us put up a PR, we could get that merged upstream?

~~Though if we were to do that, we may also want to do the analogous thing for record fields, which means not merging~~. Actually there's then seemingly nowhere good to put the haddock for the first field when using leading commas - it can't go before the `{`.

This is a draft because:
- The implementation is messy.
- I haven't tackled operators.
- This should be put behind a flag to maintain close compatibility with Ormolu. Although I'd like it to be on by default.
- The `declaration/data/operators.hs` test is failing due to an idempotency issue with `forall` and multi-line layout.
  - All the other test failures are just due to the fact that this hasn't been put behind a flag.
- As above, I'm not totally sure it's the best approach.

Interested to hear everyone's thoughts.

### DIFF
--- a/data/examples/declaration/data/fat-multiline-four-out.hs
+++ b/data/examples/declaration/data/fat-multiline-four-out.hs
@@ -2,11 +2,11 @@ module Main where
 
 -- | Something.
 data Foo
-    = -- | Foo
-      Foo
+    = Foo
+    -- ^ Foo
         Int
         Int
-    | -- | Bar
-      Bar
+    | Bar
+    -- ^ Bar
         Bool
         Bool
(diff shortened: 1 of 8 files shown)


georgefst:
I completely forgot that I ever even tried this. I guess I became demotivated because the style I'd actually prefer is supposedly hard to implement. But if someone wants to get the trailing version working instead then that'd be a great contribution. It shouldn't be too hard.

Issue #74, opened 2021-03-15
Support different comment types for module and function docstrings


brandon-leapyear:
I believe it's fairly standard for module docstrings to be of the form
```hs
{-|
Module: Foo.Bar
Description: ...
...
-}
```
and function docstrings to be of the form
```hs
-- | asdf
-- bar
-- baz
```
We should allow module docstrings to be different from function docstrings.

Also, as a bonus, it'd be great if the `{-|` syntax didn't add a space like `{- |`.

dpwiz:
> Also, as a bonus, it'd be great if the `{-|` syntax didn't add a space like `{- |`

Or did, when requested :wink:

PR #130, opened 2021-11-18
Add option for spaceless '{-|'

brandonchinn178:
Blocked on #49?
Related to #74 

It always kinda bothered me when multiline haddocks were formatted like
```hs
{- |
Some multiline
comment here
-}
```
The haddock documentation always uses `{-|` without the space

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,7 +49,7 @@ We share all bar one of Ormolu's goals:
 | `indent-wheres`      | `true`, `false` | `false` means save space by only half-indenting the `where` keyword |
 | `record-brace-space` | `true`, `false` | `rec {x = 1}` vs `rec{x = 1}` |
 | `respectful`         | `true`, `false` | Whether to respect user-specified newlines, e.g. in import groupings |
-| `haddock-style`      | `multi-line`, `single-line` | Whether multiline haddocks should use `{-` or `--` |
+| `haddock-style`      | `single-line`, `multi-line`, `multi-line-compact` | Whether multiline haddocks should use `{-` or `--` |
 | `newlines-between-decls` | any integer | number of newlines between top-level declarations |
 | `fixities`           | A list of strings for defining fixities; see the "Language extensions, dependencies, and fixities" section below |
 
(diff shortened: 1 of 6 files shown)


georgefst:
Implementation looks good, but there's more to discuss before merging this.

I personally tend to use the existing style, though I really don't care either way. Could we perhaps get stats off Hackage for how popular each of the two is (I often see these sorts of stats in GHC proposals etc. - is there an easy way to grab the source of all packages / all popular/maintained packages)? This may need to be put behind a flag.

brandonchinn178:
Yeah, I don't mind getting stats. Is there a tool people use to scrape Hackage?

georgefst:
> Is there a tool people use to scrape Hackage?

I don't know of any off-hand. I'd be interested to know if there is.

brandonchinn178:
Kind [reddit user](https://www.reddit.com/r/haskell/comments/qxrizc/how_to_scrape_hackage/) pointed me to https://hackage-search.serokell.io, which seems to work. Unfortunately, as every package uses at least one of `{- |` or `{-|`, this basically boils down to loading all packages. I'll probably just write a script that goes through some of the popular packages and summarize the data that way

brandonchinn178:
Ended up writing my own script for this: https://github.com/brandonchinn178/hackage-grep

Searching for `{-|` and `{- |` on the top 1000 packages on Hackage:
* 138 packages use just `{-|`
* 133 packages use just `{- |`
* 75 packages use both

Note that if any of those packages use `fourmolu`, then some of the `{- |` matches might be false positives (if the author would have used `{-|` if not for fourmolu). But regardless, it seems like `{- |` is surprisingly (at least to me) used fairly often, so I'll hide this behind a flag.

brandonchinn178:
@georgefst FYI this is now behind a flag. Personally, because all official documentation uses `{-|`, I would be in favor of changing the fourmolu default to `multi-line-compact`. What do you think?

georgefst:
> @georgefst FYI this is now behind a flag. Personally, because all official documentation uses `{-|`, I would be in favor of changing the fourmolu default to `multi-line-compact`. What do you think?

I'm reluctant to change defaults without good reason. There are arguments on both sides. In particular, the reason I use the current style is consistency with single-line style: `--|` is actually a parse error.

brandonchinn178:
Thanks @georgefst, I won't change the default. Good to merge, then?

georgefst (approved):

PR #133, opened 2021-11-19
Add `executable` flag to cabal file

hololeap:
This package is pulled in by the [fourmolu plugin](https://github.com/haskell/haskell-language-server/tree/master/plugins/hls-fourmolu-plugin) for `haskell-language-server`, therefore some users will not wish to build/install the executable. This gives the option to disable it, although this is not the default.

### DIFF
--- a/fourmolu.cabal
+++ b/fourmolu.cabal
@@ -113,6 +113,10 @@ library
     if impl(ghc <8.10.0)
         ghc-options: -fmax-pmcheck-iterations=3000000
 
+flag executable
+    description: Build fourmolu executable
+    default: True
+
 executable fourmolu
     main-is:          Main.hs
     hs-source-dirs:   app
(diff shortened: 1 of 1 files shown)


georgefst:
Cabal (and presumably Stack) will only ever build the library component of any packages in `build-depends`.

So I'm a bit confused. Is there a specific issue you think this is solving?

hololeap:
Right. This is a side-effect of packaging for Gentoo, which uses it's own package manager to resolve dependencies and `cabal v1-*` to build, thus everything gets built by default.

hololeap:
Since this is niche, I'll leave it up to you to decide if this PR is worth it. It isn't strictly necessary for building on Gentoo, since it is straightforward to patch the `.cabal` file, but the standard protocol is to submit patches to upstream.

hololeap:
And thanks for pointing out that Cabal does not build executables for dependencies. It explains why so many Cabal packages have extra executables which clutter up `/usr/bin/` in Gentoo.

georgefst:
Yeah, this sounds like Gentoo's approach to Haskell packaging needs a rethink, not least because v1 Cabal won't be around forever.

PR #139, opened 2021-11-30
Avoid setting Ormolu dev flag in Cabal config

georgefst:
Carrying on the conversation from https://github.com/fourmolu/fourmolu/pull/129#issuecomment-981870050:

> I actually intentionally want `-Werror` on in local development. If it's gonna fail in CI, why not have it fail locally? `cabal.project` also has it nominally on (although it should be updated to `fourmolu +dev` instead of `ormolu +dev`), so I'd rather we change it so both Stack and Cabal have `-Werror` on in local development

I've never personally understood the desire to enable `-Werror`, other than when one wants to ensure that builds output a non-zero exit code in the presence of warnings, e.g. in CI. During interactive development, surely it's often useful to run code that contains potentially-harmless issues like unused variables, or missing type signatures? I feel like I must be missing something.

Anyway, while this PR doesn't really change anything, it draws attention to the fact that we enable `-Werror` when building with Stack, but not Cabal, which is weird.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -70,7 +70,7 @@ jobs:
       # build
       - run:
           name: Build
-          command: stack build --test --no-run-tests
+          command: stack build --test --no-run-tests --flag fourmolu:dev
 
       # lint
       - when:
@@ -88,7 +88,9 @@ jobs:
       # test
       - run:
           name: Test
-          command: stack test
+          command:
+            # keep the dev flag here to avoid rebuilding from the build step
+            stack test --flag fourmolu:dev
 
       # build haddock
       - when:
--- a/cabal.project
+++ b/cabal.project
@@ -1,3 +1,1 @@
 packages: .
-
-constraints: ormolu +dev
--- a/stack-ghc-8.10.yaml
+++ b/stack-ghc-8.10.yaml
@@ -3,6 +3,3 @@ resolver: lts-18.12
 extra-deps:
   - Cabal-3.6.2.0
   - ghc-lib-parser-9.2.1.20211101
-
-ghc-options:
-  '$locals': -Werror
--- a/stack-ghc-9.0.yaml
+++ b/stack-ghc-9.0.yaml
@@ -3,6 +3,3 @@ resolver: nightly-2021-11-28
 extra-deps:
   - Cabal-3.6.2.0
   - ghc-lib-parser-9.2.1.20211101
-
-ghc-options:
-  '$locals': -Werror
--- a/stack-ghc-9.2.yaml
+++ b/stack-ghc-9.2.yaml
@@ -8,6 +8,3 @@ extra-deps:
   - base-compat-batteries-0.12.1
   - Cabal-3.6.2.0
   - ghc-lib-parser-9.2.1.20211101
-
-ghc-options:
-  '$locals': -Werror


georgefst:
I've just noticed while writing the above that #138 takes the opposite approach and sets `fourmolu +dev`. This would frustrate my local workflow.

brandonchinn178:
My preference would be to turn on the `dev` flag in both `cabal.project` and `stack.yaml`. You could always turn it off ad-hoc with `--ghc-options -Wwarn`, but I think enabling it by default is a Good Thing.

brandonchinn178:
Personally, I would be very frustrated that a working build locally with default settings fails on CI. I expect a passing `git clone` + `stack test` locally should pass CI

georgefst:
> I expect a passing `git clone` + `stack test` locally should pass CI

Ah. Well I wouldn't, in the same way that I wouldn't expect running `stack test` to check for HLint warnings, or bad formatting.

brandonchinn178:
🤷 ok, up to you. You can remove -Werror from all the stack.yaml files, then.

georgefst:
> shrug ok, up to you. You can remove -Werror from all the stack.yaml files, then.

I think I'll have to, sorry. Is it easy enough to re-enable locally to get the behaviour you want (I'm not very familiar with Stack)?

Perhaps it's worth starting an r/haskell thread or similar to find out what other people do here, and why. In my other projects, including at work, `-Werror` has only ever been enabled on CI.

georgefst:
The right course of action in that case would actually be to remove al references to `-Werror` in `fourmolu.cabal`, rather than removing the `dev` flag. Which also avoids merge conflicts with #138.

brandonchinn178:
> I think I'll have to, sorry. Is it easy enough to re-enable locally to get the behaviour you want (I'm not very familiar with Stack)?

It's easy either way. To turn it on, you just add `--ghc-options -Werror` to your `stack build` command. With it turned on by default in `stack.yaml`, you could turn it off by adding `--ghc-options -Wwarn` to your `stack build` command. Personally, I like turning it on by default because it's easier to forget to turn on than turn off.

> The right course of action in that case would actually be to remove all references to -Werror in fourmolu.cabal, rather than removing the dev flag. Which also avoids merge conflicts with #138.

No, `-Werror` in `fourmolu.cabal` is always behind the `dev` flag, so there's no point keeping it in. Plus, removing it in three locations in `fourmolu.cabal` increase the likelihood of merge conflicts than just removing the line in `cabal.project` (plus, `fourmolu.cabal` will be touched more often upstream than `cabal.project`)

georgefst:
> No, `-Werror` in `fourmolu.cabal` is always behind the `dev` flag, so there's no point keeping it in. Plus, removing it in three locations in `fourmolu.cabal` increase the likelihood of merge conflicts than just removing the line in `cabal.project` (plus, `fourmolu.cabal` will be touched more often upstream than `cabal.project`)

Well, on the other hand, for the sake of argument:
- We'd then miss out on some extra warnings (but admittedly nothing terribly important - `-Wincomplete-uni-patterns` is even on by default since GHC 9.2, although somehow this didn't make it in to the release notes...)
- Those lines are unlikely to change at all often.

But I don't really care much either way, so I'll do it your way!

brandonchinn178:
Well either way, we'll need to make sure to add the dev flag on CI

brandonchinn178:
Rebase against master and resolved comments according to discussion:
* Keep `dev` flag + `-Werror` off by default with both cabal and stack
* Keep `dev` flag + `-Werror` in cabal file (avoids conflicts with upstream)
* Add `dev` flag + `-Werror` to CI

To reiterate, I personally think dev/Werror should be on by default, since it's easy to turn off if you make changes and want to temporarily turn off warnings (both cabal + stack support `--ghc-options=-Wwarn` to the build command), and it's easier to remember to turn off errors than to remember to turn them on. But I'd much rather we merge something than leave things as-is, hence the changes above.

To try to reduce number of open PRs, I'm merging this as soon as CI passes. If changes are desired, we can open up another PR.

PR #144, opened 2022-01-10
Add Brandon to maintainers list in Cabal file

georgefst:
Should have happened when we gave him permission. Just an oversight.


brandonchinn178 (approved):
Thanks!

PR #149, opened 2022-01-23
Fix indentation after list comprehension

### LINKED ISSUE #142: Regression with 0.5.0.0: Fix indentation after a list comprehension with infix operator
Fourmolu unexpectedly changes
```hs
x =
  [1 | _ <- _]
    <> [2]
```
to
```hs
x =
  [1 | _ <- _]
  <> [2]
```
This only seems to happen with list comprehensions. This occurred when merging Ormolu 0.4.0.0; Ormolu has this correct

https://github.com/fourmolu/fourmolu/pull/138#discussion_r758708927

brandonchinn178:
Fixes #142 

Ormolu refactored all this code after 0.4.0.0 was released, so this code will probably go away. But fixes the weirdness for now.

Ref. https://github.com/tweag/ormolu/pull/830

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 * Fixed regression in 0.5.0.0 with multiline tuples ([#162](https://github.com/fourmolu/fourmolu/pull/162))
 * Fixed regression in 0.5.0.0 with multiline record types ([#160](https://github.com/fourmolu/fourmolu/pull/160))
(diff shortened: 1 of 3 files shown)


georgefst (approved):

georgefst:
Looks good. I guess we should have a regression test, especially if it might slip back in when we merge upstream. But happy to delay that until we've decided what we're doing about testing in general.

Issue #152, opened 2022-02-09
Allow configuring indentation of "in" after "let" block


bapcyk:
![image](https://user-images.githubusercontent.com/14261219/153221006-7db4be6d-0bbd-4a82-8b51-5bfca07c78de.png)

`in` is shifted to 1 space, how to shift it to indentation * N, where N=0, 1, ... ? TBH 0 would be better :) Is it possible? Because currently it is not multiple of indentation.

brandonchinn178:
That's not currently configurable. But yes, we could add that configuration

brandonchinn178:
Also related: #83

PR #135, opened 2021-11-23
Add haddock-style-module configuration

brandonchinn178:
Blocked on #49?
Resolves https://github.com/fourmolu/fourmolu/issues/74

Adds a `haddock-style-module` option that specifies the haddock style to use specifically for the module docstring (when not set, resolves to whatever `haddock-style` is — the current behavior)

Tested with this file:
```hs
{- |
Module: Foo
-}

module Foo where

-- | This is a
-- multi line
-- haddock
foo :: Int
foo = 42
```
and this command:
```console
$ stack exec -- fourmolu --haddock-style=single-line --haddock-style-module=multi-line Foo.hs

{- |
Module: Foo
-}
module Foo where

-- | This is a
-- multi line
-- haddock
foo :: Int
foo = 42
```

Issue #84, opened 2021-04-25
Consistent handling of `do`, `let`, `where`


AshleyYakeley:
There are a number of possible ways of handling these keywords:

1. sameline/newline
```haskell
alpha = beta where
    gamma = let
        delta = do
            epsilon
        in zeta
```
2. newline-indent/newline-indent
```haskell
alpha = beta
    where
        gamma =
            let
                delta =
                    do
                        epsilon
            in zeta
```
3. newline-half/newline-half
```haskell
alpha = beta
  where
    gamma =
      let
        delta =
          do
            epsilon
      in zeta
```
4. newline-indent/newline
```haskell
alpha = beta
    where
    gamma =
        let
        delta =
            do
            epsilon
        in zeta
```
5. newline-indent/align
```haskell
alpha = beta
    where gamma1 = undefined
          gamma2 =
              let delta1 = undefined
                  delta2 =
                      do epsilon1
                         epsilon2
              in zeta
```
Currently the default behaviour seems to be:
* `do`: sameline/newline
* `let`: newline-indent/align
* `where`: newline-half/newline-half (can be set to newline-indent/newline-indent)

It would be great if there were settings to specify the behaviour of each of these keywords.

brandonchinn178:
As discussed in #155, I don't agree with having a single option that makes all of these consistent. It's also not clear to me why these three constructs, are being discussed in particular. Why not `if`? `let` and `where` both define bindings; how is `do` relevant to how `let` or `where` is styled?

But I would be open to incrementally changing the styles of each of these constructs separately, to allow configuring how `do` blocks are styled / configuring how `let` blocks are styled / etc.

I'm closing this issue, because it doesn't make sense to me to tie these three constructs together. But feel free to open new issues around how `do` and `where` should be configurable (`let` styling will be discussed here: https://github.com/fourmolu/fourmolu/issues/202)

PR #77, opened 2021-03-18
Add option for inserting newline after a 'let'

georgefst:

PR #155, opened 2022-02-14
"Consistent Let" option

AshleyYakeley:
This creates a new configuration option, `consistent-let`. When false, the existing ugly Ormolu layout is used for `let` statements:
```
e1 =
    let a = 1
        b = 2
     in c

e2 = do
    a
    let b = 1
        c = 2
    return d
```
When true, Fourmolu treats `let` consistently with `do`, `case`, and `\case`, allowing the keyword to hang, and indenting the bindings & body normally:
```
e1 = let
    a = 1
    b = 2
    in c

e2 = do
    a
    let
        b = 1
        c = 2
    return d
```
Note that this is still allowed, if there's only one binding:
```
e2 = do
    a
    let b = 1
    return d
```

### DIFF
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -384,6 +384,14 @@ printerOptsParser = do
           "Number of spaces between top-level declarations"
             <> showDefaultValue poNewlinesBetweenDecls
       ]
+  poConsistentLet <-
+    (optional . option parseBoundedEnum . mconcat)
+      [ long "consistent-let",
+        metavar "BOOL",
+        help $
+          "Whether to lay out 'let' similar to 'do', 'case', and '\\case'"
+            <> showDefaultValue poConsistentLet
+      ]
   pure PrinterOpts {..}
 
 sourceTypeParser :: Parser (Maybe SourceType)
--- a/src/Ormolu/Config.hs
+++ b/src/Ormolu/Config.hs
@@ -137,7 +137,9 @@ data PrinterOpts f = PrinterOpts
     -- | How to print doc comments
     poHaddockStyle :: f HaddockPrintStyle,
     -- | Number of newlines between top-level decls
-    poNewlinesBetweenDecls :: f Int
+    poNewlinesBetweenDecls :: f Int,
+    -- | Whether to lay out let the same as do, case, and lambda-case
+    poConsistentLet :: f Bool
   }
   deriving (Generic)
 
@@ -153,7 +155,7 @@ instance Semigroup PrinterOptsPartial where
   (<>) = fillMissingPrinterOpts
 
 instance Monoid PrinterOptsPartial where
-  mempty = PrinterOpts Nothing Nothing Nothing Nothing Nothing Nothing Nothing Nothing
+  mempty = PrinterOpts Nothing Nothing Nothing Nothing Nothing Nothing Nothing Nothing Nothing
 
 -- | A version of 'PrinterOpts' without empty fields.
 type PrinterOptsTotal = PrinterOpts Identity
@@ -172,7 +174,8 @@ defaultPrinterOpts =
       poDiffFriendlyImportExport = pure True,
       poRespectful = pure True,
       poHaddockStyle = pure HaddockMultiLine,
-      poNewlinesBetweenDecls = pure 1
+      poNewlinesBetweenDecls = pure 1,
+      poConsistentLet = pure False
     }
 
 -- | Fill the field values that are 'Nothing' in the first argument
@@ -192,7 +195,8 @@ fillMissingPrinterOpts p1 p2 =
       poDiffFriendlyImportExport = fillField poDiffFriendlyImportExport,
       poRespectful = fillField poRespectful,
       poHaddockStyle = fillField poHaddockStyle,
-      poNewlinesBetweenDecls = fillField poNewlinesBetweenDecls
+      poNewlinesBetweenDecls = fillField poNewlinesBetweenDecls,
+      poConsistentLet = fillField poConsistentLet
     }
   where
     fillField :: (forall g. PrinterOpts g -> g a) -> f a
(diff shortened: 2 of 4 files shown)


AshleyYakeley:
Relevant: #76, #77, #84.

brandonchinn178:
👋 @AshleyYakeley 
Thank you for opening this PR and being patient while we sorted out how to handle all the new contributions for new configuration options!

If you haven't noticed, we finally merged in a new testing framework for configuration options ([#169](https://github.com/fourmolu/fourmolu/pull/169))! This means we can finally address the backlog of open PRs.

Please do the following steps

- [ ] Rebase against `main`

    ```bash
    git remote add upstream git@github.com:fourmolu/fourmolu
    git fetch upstream
    git rebase -i upstream/main
    ```

- [ ] Resolve merge conflicts
- [ ] Ensure `CHANGELOG.md` has been updated
- [ ] Ensure configuration docs in `README.md` have been updated
- [ ] Ensure `fourmolu.yaml` updated to stay in sync with config in `README.md`
- [ ] Ensure tests have been added

You can use [#135](https://github.com/fourmolu/fourmolu/pull/135) as a reference for this checklist.

AshleyYakeley:
@brandonchinn178 Is this PR likely to be merged, in the event that I complete these steps and there are no problems?

brandonchinn178:
Can you explain the difference the approach here and the approach in #77? To me, #77 seems to be solving the same problem, and is slightly better for the following reasons:
1. Easily explainable. I had to read a lot of the linked issues in this PR to understand why let is "inconsistent", and what "being consistent with case and where" means. The other PR is much easier to immediately grok: just adds a newline before let. The config option is, thus, much more understandable: "let-newline" vs "consistent-let"
2. The other PR seems to be slightly more contained. When updating this PR, please make sure to change as little as possible (see DEVELOPER.md regarding refactoring and upstream changes)
3. The other PR's implementation is more value-neutral; I would hesitate to have functions in the project named "ugly style" vs "consistent style"

I am probably missing something here, though, so please enlighten me!

AshleyYakeley:
For some reason Ormulu (and therefore fourmolu) treats `let` differently from `do`, `case`, and `\case`. This option makes it consistent. So obviously this is a bigger change than #77, but it ends up with more consistent behaviour.

Have a look at these:
```
l1 = let
    a = 1
    b = 2
    in c

d1 = do
    a
    b
    return c

l2 =
    let a = 1
        b = 2
     in c

d2 =
    do a
       b
       return c
```

Can you see how `l1` and `d1` look the same? And `l2` and `d2` also kinda look the same? This is what I mean by consistency. But right now fourmolu inconsistently does `l2` and `d1`.

brandonchinn178:
> > I would argue that this is desired behavior; to me (and, in my experience, to most other Haskell devs), `let` occupies a different semantic space than `do` and `case`.
> 
> This seems like merely a justification for the inconsistency?

No, I don't mean that `let` is one semantic group and `do` + `case` are in a second semantic group. I mean they all occupy separate semantic spaces, and thus any similarities or differences between their styles is purely coincidental. My point here is that there's not much/any value making things "consistent" between these semantic groups. If you want X change to make `let`/`do`/`case` styled closer to what you want, great, but I think that X should be judged solely within the context of `let`/`do`/`case`, and not within any context of consistency with any of the other ones.

> > this proposal seems to me like consistency-for-the-sake-of-consistency
> 
> This is exactly what this proposal is! I want the style of my code to look as consistent and predictable and simple as possible.

My point here is that because all of these keywords are in different semantic groups from each other, talking about "consistency" is completely absurd to me. To me, trying to link `let` and `do` is akin to linking `let` and `module`; just because they're keywords, doesn't mean that there's any value to making them consistent.

Of course, all of this is completely my personal opinion. I understand if you don't want to do the work of updating the PR if it's unknown if it'll be accepted. I'm fine keeping this open (or closed) until we hear more feedback from the community (either in this PR, in #84, or some other issue).

But what I would be open to is incremental changes to get to the eventual goal you want, e.g.

> For example, I want no off-multiple indents ever (e.g. the 5 spaces before `in` in `l2`).

This is captured by #152 and partially captured by https://github.com/fourmolu/fourmolu/pull/83. I think I'll open up a new issue to discuss `let`/`in` styling.

AshleyYakeley:
> My point here is that because all of these keywords are in different semantic groups from each other, talking about "consistency" is completely absurd to me. To me, trying to link `let` and `do` is akin to linking `let` and `module`; just because they're keywords, doesn't mean that there's any value to making them consistent.

Isn't consistency the whole point of styling code? One should set a minimal number of design principles, and make all constructions follow them as cleanly as possible.

In this case, if you know how one or two of `let`, `do`, and `case` are styled, you should be able to guess how the others are. This makes it easier and more pleasant to read IMO.

brandonchinn178:
> Isn't consistency the whole point of styling code? One should set a minimal number of design principles, and make all constructions follow them as cleanly as possible.

Sure, consistency in that all `if` statements are styled the same, all data constructors are styled the same, all `do` blocks are styled the same, etc. But I don't think there's any expectation for `let` and `do` to share styling, just because they're both keywords, in the same way you wouldn't expect `let` to have the same styling as `module` or `if`.

> In this case, if you know how one or two of `let`, `do`, and `case` are styled, you should be able to guess how the others are.

What's the need for guessing? You just write what you write, then the formatter corrects it.

> This makes it easier and more pleasant to read IMO.

🤷 IMO not really.

I wrote a comment in #84 and closed it in favor of writing new issues specific to each type of construct. I'm going to close this PR for the same reason as closing that issue; I don't think this configuration option goes in the right direction. But I look forward to any future issues + PRs around this area!

AshleyYakeley:
> But I don't think there's any expectation for `let` and `do` to share styling, just because they're both keywords, in the same way you wouldn't expect let to have the same styling as `module` or `if`.

TBH I think there's an expectation for the code to look like it follows consistent principles. Of course, there are always trade-offs, but where things can look the same they probably ought to. For example, `if` is comparable to `case` and should probably follow a comparable style, where appropriate. `class` and `instance` are similar constructions, and I think most people would be very surprised if the styles for the two were entirely unrelated.

Since there have already been a number of issues in this area, I think at this point it's up to you to determine what the right direction is, and then implement it, rather than waiting for more issues/PRs. If you happen to decide to make everything highly configurable, I would be perfectly happy if I can configure it just the way I want.

Issue #145, opened 2022-01-10
Regression: fourmolu-0.5.0.0 deindents multiline types in record fields


brandonchinn178:
Fourmolu reformats this:
```hs
newtype App a = App
  { unApp ::
      ReaderT
        MyConfig
        (LoggingT IO)
        a
  }
```
as
```hs

newtype App a = App
  { unApp ::
    ReaderT
      MyConfig
      (LoggingT IO)
      a
  }
```

Possibly caused by https://github.com/fourmolu/fourmolu/pull/124

PR #160, opened 2022-03-16
Fix regression with multiline record types

brandonchinn178:
Partially revert https://github.com/fourmolu/fourmolu/pull/124 (specifically, the removal of `sitcc`) to fix https://github.com/fourmolu/fourmolu/issues/145

Completely separate from whether it looks better or not, that change actually causes inconsistencies with different options. Some examples are below. TL;DR we should revert this for now to fix the inconsistencies, then we can make the trailing comment indentation configurable for aesthetic reasons

## Record field inconsistencies

With `--indentation=4`, `--comma-style=trailing` and `--comma-style=leading` will indent the two comments differently:

```console
$ stack exec -- fourmolu --indentation=4 --comma-style=trailing test.hs
newtype App a = App
    { unApp ::
        JobExecutorT
            ( ReaderT
                MyConfig
                (LoggingT IO)
            )
            a,
      -- | asfd
      a :: Int,
      -- | asdf
      b :: Bool
    }

$ stack exec -- fourmolu --indentation=4 --comma-style=leading test.hs
newtype App a = App
    { unApp ::
        JobExecutorT
            ( ReaderT
                MyConfig
                (LoggingT IO)
            )
            a
    , a :: Int
    -- ^ asfd
    , b :: Bool
    -- ^ asdf
    }
```
But _**in addition to this**_, using `--indentation=2`, it also causes multiline types to be deindented (note `JobExecutorT` being deindented after `unApp` with `--comma-style=leading`):
```console
$ stack exec -- fourmolu --indentation=2 --comma-style=trailing test.hs
newtype App a = App
  { unApp ::
      JobExecutorT
        ( ReaderT
            MyConfig
            (LoggingT IO)
        )
        a,
    -- | asfd
    a :: Int,
    -- | asdf
    b :: Bool
  }

$ stack exec -- fourmolu --indentation=2 --comma-style=leading test.hs
newtype App a = App
  { unApp ::
    JobExecutorT
      ( ReaderT
          MyConfig
          (LoggingT IO)
      )
      a
  , a :: Int
  -- ^ asfd
  , b :: Bool
  -- ^ asdf
  }
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## Unreleased
 
 * Fixed regression in 0.5.0.0 with multiline tuples ([#162](https://github.com/fourmolu/fourmolu/pull/162))
+* Fixed regression in 0.5.0.0 with multiline record types ([#160](https://github.com/fourmolu/fourmolu/pull/160))
 * Add `import-export-comma-style` configuration to allow leading commas in module import/export lists
 
 ## Fourmolu 0.5.0.0
--- a/src/Ormolu/Printer/Meat/Type.hs
+++ b/src/Ormolu/Printer/Meat/Type.hs
@@ -259,7 +259,7 @@ p_forallBndrs vis p tyvars =
 
 p_conDeclFields :: [LConDeclField GhcPs] -> R ()
 p_conDeclFields xs =
-  braces N $ sep commaDel (located' p_conDeclField) xs
+  braces N $ sep commaDel (sitcc . located' p_conDeclField) xs
 
 p_conDeclField :: ConDeclField GhcPs -> R ()
 p_conDeclField ConDeclField {..} = do
@@ -276,7 +276,7 @@ p_conDeclField ConDeclField {..} = do
   breakpoint
   sitcc . inci $ p_hsType (unLoc cd_fld_type)
   when (commaStyle == Leading) $
-    mapM_ ((newline >>) . p_hsDocString Caret False) cd_fld_doc
+    mapM_ (inciByFrac (-1) . (newline >>) . p_hsDocString Caret False) cd_fld_doc
 
 tyOpTree :: LHsType GhcPs -> OpTree (LHsType GhcPs) (LocatedN RdrName)
 tyOpTree (L _ (HsOpTy _ l op r)) =


brandonchinn178:
@georgefst it would be great to merge this + release 0.5.1 to fix the regression quickly

georgefst:
I'm reluctant to merge this _as-is_, since unfortunately while it fixes one regression, in a sense it introduces a new one. I, for one, am already using Fourmolu 0.5 and would rather avoid the churn that this would entail. I also don't quite understand all of your concerns, so I could do with a little explanation.

> With --indentation=4, --comma-style=trailing and --comma-style=leading will indent the two comments differently:

I don't really see why this is an issue? Why should different options be consistent in this way?

> But in addition to this, using --indentation=2, it also causes multiline types to be deindented

That certainly _is_ problematic.

I haven't tested this extensively, but I think we can keep the best of both worlds if we hack in an indentation decrease by changing line 279 to:
```hs
mapM_ (inciBy (-2) . (newline >>) . p_hsDocString Caret False) cd_fld_doc
```

> It also shows up with multiline tuples:

I'm unsure what I'm supposed to be looking at here. I think the first output is a bit ugly, but that isn't affected by removing or re-adding `sitcc`.

brandonchinn178:
> > With --indentation=4, --comma-style=trailing and --comma-style=leading will indent the two comments differently:
> 
> I don't really see why this is an issue? Why should different options be consistent in this way?

I guess, personally, I would expect that the comments, regardless of being `-- |` or `-- ^`, should always be either at the same indentation level as the initial `{` or indented past `{`. But I guess I do recognize this is more on the subjective end, so I won't push much more here.

> > But in addition to this, using --indentation=2, it also causes multiline types to be deindented
> 
> That certainly _is_ problematic.
> 
> I haven't tested this extensively, but I think we can keep the best of both worlds if we hack in an indentation decrease by changing line 279 to:
> 
> ```haskell
> mapM_ (inciBy (-2) . (newline >>) . p_hsDocString Caret False) cd_fld_doc
> ```

I can try that

> > It also shows up with multiline tuples:
> 
> I'm unsure what I'm supposed to be looking at here. I think the first output is a bit ugly, but that isn't affected by removing or re-adding `sitcc`.

If you look at the first example, the do block is not indented in the second element of the tuple. it's indented in every other example

georgefst:
> If you look at the first example, the do block is not indented in the second element of the tuple. it's indented in every other example

Sure, but this output isn't affected by the changes in #124. It appears to be a separate issue. Is it a regression? If so, we should try to bisect.

brandonchinn178:
ah you're right... i'm not sure why i connected that issue to this. i'll move that discussion to a separate branch

brandonchinn178:
@georgefst Fixed! Tested that the below does not reformat with `--indentation=2 --comma-style=leading`
```hs
newtype App a = App
  { unApp ::
      JobExecutorT
        ( ReaderT
            MyConfig
            (LoggingT IO)
        )
        a
  , a :: Int
  -- ^ asfd
  , b :: Bool
  -- ^ asdf
  }
```
I didn't add this as a test yet, because I'm not sure what we want to do about testing (did you get my latest email?). I'm personally of the mindset that `data/examples/` should ONLY contain ormolu test cases, and any test cases fourmolu adds should be in a separate directory like `data/examples-fourmolu/`, to make it clear what tests are upstream and what tests we added.

brandonchinn178:
@georgefst Thanks! I'll be making a release shortly

PR #169, opened 2022-04-16
Add test framework for Fourmolu options

### LINKED ISSUE #49: Testing all options

### LINKED ISSUE #56: Confusing docs for `diff-friendly-import-export`.
The docs for this option say "Whether to make use of extra commas in import/export lists (as opposed to Ormolu's style)", but Ormolu uses trailing commas on multi-line import and export lists (and I'm pretty sure it has since before fourmolu was forked), which would seem to be the "diff friendly" style.

In #19, tweag/ormolu#409 is referenced, but that issue isn't about commas, but rather the position of the parens.

So, I guess I'm asking for that documentation to be more explicit about how that option actually affects the formatting.

brandonchinn178:
Resolves #49, Resolves #56

- [x] Revert `PrinterSpec.hs` + `data/` to basically what it is in Ormolu
- [x] Add tests for current options
- [x] Add checklist to docs for adding a new config option
- [x] Add to docs list of current options


brandonchinn178:
cc @georgefst Thoughts on both the test strategy + the input files? Did I cover the main styling differences for each option?

arybczak:
Hey. Is this PR blocked by anything?

brandonchinn178:
Hey @arybczak! I was hoping to get @georgefst to take a look, but he seems to be busy. I'm actually going on vacation, so I was going to merge this when I get back, give George a little more time anyway

brandonchinn178:
FYI @3kyro this reverts the tests you added in #108, with new tests captured in https://github.com/fourmolu/fourmolu/pull/169/files#diff-2ce4cd9cc4e15f9f8fdd33721a2b90f5eade9582ca70bc09e6af9a9928fc5346. Can you check to see if there are any other cases we should test here?

@georgefst To get PRs unblocked and get activity flowing again in this repo, I'm going to merge this soon. It'd be great to get your review still, whether it comes before or after I merge.

data/fourmolu/import-export/output-Leading.hs
```diff
@@ -1,3 +1,10 @@
+module Foo
+    ( foo
+    , bar
+    , baz
+    )
+where
+
 import qualified MegaModule as M
```
3kyro, line 8: I think we should add a multi line import (or export) to test that commas respect the import-export list option (as per #187). It's probably best to check alignment of section headings and comments as well.
````haskell
module My.Module
    ( -- * Something
      foo
    , bar
      -- | A multiline
      -- comment here
    , bar2

      -- * Another thing
    , (<?>)
    )

import qualified MegaModule as M
    ( Either
    , Monad
        ( return
        , (>>)
        , (>>=)
        )
    , (<<<)
    , (>>>)
    )
````

brandonchinn178:
@3kyro thanks! Updated here: https://github.com/fourmolu/fourmolu/pull/169/commits/1f80c2d148b76127c08f0a220daad5cf03f3cd57

The style's a bit wonky, but we can leave that off for later. I'll merge this in when CI passes

3kyro:
> The style's a bit wonky, but we can leave that off for later. I'll merge this in when CI passes

@brandonchinn178  Yes, styling with leading and diff-friendly is terrible. A ternary option would be great to have  but we can start by ignoring the diff-friendly setting when import export commas are set to leading. Let me know if there's anything I can do.

brandonchinn178:
@3kyro ah yes ok, I did mention that here: https://github.com/fourmolu/fourmolu/pull/108#issuecomment-1100916957. Nothing to do now. Maybe we can take a look after I release the next version

PR #170, opened 2022-04-17
Add --quiet flag

### LINKED ISSUE #110: Disable output of "Loaded config from: <path>"?
When loading a custom configuration in `fourmolu.yaml`, the above string is written to `stderr`. I'd like to be able to disable this (maybe via CLI option), as it interferes with my tooling - even though the tooling should in my opinion also be able to handle this better. The following is what I'm having problems with:

`haskell-mode` for Emacs has a configuration option to set the formatter, which I set to `fourmolu`. It then starts a background process, capturing `stdout` and in the success case, it overwrites the buffer contents with that. But it also captures `stderr` and in the case that the corresponding file is non-empty, it fails altogether and won't format the buffer. I think it's worth for me to try and fix that upstream by checking the error code with precedence over the `stderr` file, but it's maybe also worth making it configurable in `fourmolu` (or something else entirely).

Or maybe not writing to `stderr` in the success case is the right call? The effects of introducing/changing a configuration file should be immediately visible or at least easily observable.

In case you think something could/should be done here, kindly let me know and I'll see if I can make a PR out of it :)

brandonchinn178:
Supercedes #101, fixes #110

Solves the immediate issue of suppressing the "Loaded config from: ..." output. Long-term, we should probably think about whether we want to consolidate `optQuiet` and `cfgDebug` into a single "log level" field (but keeping them separate cli flags); but maybe we want to keep ormolu-owned logging separate from fourmolu-owned logging?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## Unreleased
+
+* Add `--quiet` flag to make the output quieter
+
 ## Fourmolu 0.6.0.0
 
 * Fixed regression in 0.5.0.0 with multiline tuples ([#162](https://github.com/fourmolu/fourmolu/pull/162))
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -164,6 +164,8 @@ formatOne CabalDefaultExtensionsOpts {..} mode reqSourceType rawConfig mpath =
 data Opts = Opts
   { -- | Mode of operation
     optMode :: !Mode,
+    -- | Whether to make the output quieter
+    optQuiet :: !Bool,
     -- | Ormolu 'Config'
     optConfig :: !(Config RegionIndices),
     -- | Options for respecting default-extensions from .cabal files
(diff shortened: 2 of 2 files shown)


georgefst:
> but maybe we want to keep ormolu-owned logging separate from fourmolu-owned logging?

I wouldn't. The end user mostly doesn't care about the distinction.

brandonchinn178:
I would have to see a concrete example to see how keeping them separate would affect the end user. My motivation for keeping them separate is to avoid conflicts.

georgefst:
> I would have to see a concrete example to see how keeping them separate would affect the end user.

I just meant that having separate flags for such similar functionality _might_ be confusing. But I'm not that concerned about it right now.

brandonchinn178:
oh sure, I was mostly talking about how theyre stored in the Config type. I think as a user, it makes sense to have distinct --quiet and --debug/--verbose flags, but I'm not opposed to unifying those!

PR #177, opened 2022-05-14
Improve region tests

brandonchinn178:
The test suite doesn't import `fourmolu`, so it doesn't need it as a dependency. However, running tests does require the executable, so one needs to remember to explicitly build the executable. Keeping `fourmolu` as a dependency doesn't help with this; the dependency only specifies the library as a dependency, not the executable, and Stack will only build the executable the first time.

Also show better output if the command fails

### DIFF
--- a/DEVELOPER.md
+++ b/DEVELOPER.md
@@ -115,10 +115,9 @@ Fourmolu aims to continue merging upstream changes in Ormolu. Whenever Ormolu ma
 1. Add Ormolu as an upstream remote: `git remote add ormolu git@github.com:tweag/ormolu`
 1. Check out a new branch: `git switch -c merge-ormolu`
 1. Pull Ormolu's git history: `git fetch ormolu --no-tags`
-1. Find the commit corresponding to the new Ormolu version and merge it: `git merge <commit>`
+1. Find the commit corresponding to the new Ormolu version and merge it: `git merge <commit> -m 'Merge ormolu-X.Y.Z'`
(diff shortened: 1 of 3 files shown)


georgefst:
> The test suite doesn't import `fourmolu`, so it doesn't need it as a dependency

Can't we add it under `build-tool-depends`? I'm a bit concerned that it would be too easy for tests to erroneously pass locally, due to an unexpected globally-installed Fourmolu binary being used. Though presumably CI would catch this?

brandonchinn178:
hm that does solve the problem of the test suite only building the executable the first time, but it's also lying, since the exe isnt actually used at runtime. Alternatively, we might be able to add Paths_fourmolu to the test suite modules, and use `bindir` from there, which solves the problem of finding the right executable, but not the problem of building the exe each time. Or both...

georgefst:
> but it's also lying, since the exe isnt actually used at runtime

Isn't it? What do you mean?

brandonchinn178:
Sorry, build time*

georgefst:
Ah, right. I don't think that's an issue. Using it for a test suite is [encouraged by the official docs](https://cabal.readthedocs.io/en/3.6/cabal-package.html?highlight=build-tool-depends#pkg-field-build-tool-depends). In which case, you could argue the field is badly named...

brandonchinn178:
Oh that's a terribly named field... haha. Then yes, it should be included by build-tool-depends. Feel free to do so, or I can do it when I get back from vacation

brandonchinn178:
https://github.com/fourmolu/fourmolu/pull/197

PR #179, opened 2022-05-15
Clean up redundant test files

brandonchinn178:
Forgot to do this step in #175


georgefst:
Sorry, I suspect I'm being dense here, but why are these redundant?

brandonchinn178:
Because ormolu removed the source files for these upstream. So something like

1. Ormolu version X has `test.hs` and corresponding `test-out.hs`
2. Fourmolu version X generated `test-four-out.hs`
3. Ormolu version X+1 removed `test.hs` and `test-out.hs`
4. Fourmolu version X+1 has redundant file

This is documented in DEVELOPER.md; I could probably add the same command to CI to check for this...

georgefst:
Aha, thanks.

> This is documented in DEVELOPER.md; I could probably add the same command to CI to check for this...

Would be nice I suppose, but not very high priority. Seeing as having them around doesn't break anything. And any old ones will be caught by the next person who _does_ remember.

PR #180, opened 2022-05-15
Use better YAML library

brandonchinn178:
`yaml` is more standard + has better error messages than `HsYAML`:
```yaml
# fourmolu.yaml
foo: 1
bar
```
```console
# HsYAML
Failed to load /Users/brandino/repos/fourmolu/fourmolu.yaml:
  Unexpected 'b'
```
```console
# yaml
Failed to load /Users/brandino/repos/fourmolu/fourmolu.yaml:
YAML parse exception at line 2, column 0,
while scanning a simple key:
could not find expected ':'
```
Plus some clean up commits

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,9 +83,11 @@ $ stack install fourmolu
 ## Building from source
 
 ```console
-$ cabal build
-$ stack build
+$ cabal build -fdev
+$ stack build --flag fourmolu:dev
 ```
+
+The `dev` flag may be omitted in your local workflow as you work, but CI may not pass if you only build without the `dev` flag.
 
 ## Usage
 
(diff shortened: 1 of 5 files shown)


georgefst:
Foolishly, I never put my reasons for switching to `HsYaml` in the commit message of https://github.com/fourmolu/fourmolu/commit/32f85d23c9106816e50bd198857e3273ac3d6bd8. That dates from the early days when this repo was managed a bit less professionally...

I suspect it may have had to do with the GPL license.

brandonchinn178:
Thanks, thats helpful! Do you have a preference? I don't know too much about licenses; why would GPL be preferable over BSD?

georgefst:
Argh, sorry, I'm tired, and had the libraries the wrong way round! GPL is problematic, as in theory it forces us to license Fourmolu as GPL (although other reverse dependencies of `HsYAML` don't seem to have done so...). So it's good that we're moving away from `HsYAML`!

(I am curious about why I made the switch, but I guess that's now lost in the mists of time. It probably doesn't matter.)

PR #203, opened 2022-06-06
Check for constructor duplicates in `mkBijectiveMap`

3kyro:
Constructors should not be provided multiple times when building a `BijectiveMap` using `mkBijectiveMap`

cc @brandonchinn178


brandonchinn178 (approved):
Thanks!

PR #206, opened 2022-06-18
Add option to format with Unicode syntax.

### LINKED ISSUE #134: Preserve Unicode Syntax
I'd like fourmolu to preserve my beautiful ∀s. 

I think it should either get a config option like `respect-unicode-syntax`, it should leave ∀s always alone or the behavior could be made dependent on the option `respectful`.

kindaro:
Resolves #134 

Unicode looks delicious _(with an appropriate font)_, but there is no input method for it out of the box. Therefore, most people avoid it. Now we do not need an input method — it will input itself, solving the problem for good.

### DIFF
--- a/data/fourmolu/unicode-syntax/output-UnicodeAlways.hs
+++ b/data/fourmolu/unicode-syntax/output-UnicodeAlways.hs
@@ -1,0 +1,94 @@
+{-# LANGUAGE Arrows #-}
+{-# LANGUAGE DataKinds #-}
+{-# LANGUAGE GADTs #-}
+{-# LANGUAGE LinearTypes #-}
+{-# LANGUAGE PatternSynonyms #-}
+{-# LANGUAGE RankNTypes #-}
+{-# LANGUAGE TemplateHaskell #-}
+{-# LANGUAGE TypeFamilies #-}
+{-# LANGUAGE TypeFamilyDependencies #-}
+{-# LANGUAGE UnicodeSyntax #-}
+
+module UnicodeExample where
+
+import Control.Arrow
+import Data.Kind
+import Language.Haskell.TH (Exp, Quote)
+import Prelude hiding (elem)
+
+data T = MkT {foo ∷ Int}
+
+data Term a where
+    Lit ∷ Int → Term Int
+    Succ ∷ Term Int → Term Int
+    IsZero ∷ Term Int → Term Bool
+    If ∷ Term Bool → Term a → Term a → Term a
+    Pair ∷ Term a → Term b → Term (a, b)
+
+newtype Swizzle = MkSwizzle (∀ a. Ord a ⇒ [a] → [a])
+
+data family GMap k ∷ Type → Type
+
+data instance GMap (Either a b) v = GMapEither (GMap a v) (GMap b v)
+
+type family F a where
+    F Int = Double
+    F Bool = Char
+    F a = String
+
+type family Id a = r | r → a
+
+type instance Id Int = Int
+
+type instance Id Bool = Bool
+
+type family Elem c ∷ Type
+
+type instance Elem [e] = e
+
+data T1 a = MkT1 a
+
+construct ∷ a ⊸ T1 a
+construct x = MkT1 x
+
+deconstruct ∷ T1 a ⊸ a
+deconstruct (MkT1 x) = x
+
+pattern HeadC x ← x : xs
+    where
+        HeadC x = [x]
+
+pattern Point ∷ Int → Int → (Int, Int)
+pattern Point{x, y} = (x, y)
+
+class (Monad m, Monad (t m)) ⇒ Transform t m where
+    lift ∷ m a → (t m) a
+
+data Tree a = Leaf a | Branch (Tree a) (Tree a)
+
+instance (Eq a) ⇒ Eq (Tree a) where
+    Leaf a == Leaf b = a == b
+    (Branch l1 r1) == (Branch l2 r2) = (l1 == l2) && (r1 == r2)
+    _ == _ = False
+
+add1 ∷ Quote m ⇒ Int → m Exp
+add1 x = [|x + 1|]
+
+monad = do
+    putStr "x: "
+    l ← getLine
+    return (words l)
+
+arrow f g h = proc x → do
+    y ← f ⤙ x + 1
+    g ⤙ 2 * y
+    let z = x + y
+    t ← h ⤙ x * z
+    returnA ⤙ t + z
+
+elem ∷ (Eq a) ⇒ a → [a] → Bool
+x `elem` [] = False
+x `elem` (y : ys) = x == y || (x `elem` ys)
+
+h ∷ (∀ a. a → a) → (Bool, Char)
+h f = (f True, f 'c')
--- a/src/Ormolu/Config.hs
+++ b/src/Ormolu/Config.hs
@@ -230,6 +230,7 @@ overFieldsM f $(unpackFieldsWithSuffix 'PrinterOpts "0") = do
   poLetStyle <- f poLetStyle0
   poInStyle <- f poInStyle0
   poRespectful <- f poRespectful0
+  poUnicode <- f poUnicode0
   return PrinterOpts {..}
 
 defaultPrinterOpts :: PrinterOptsTotal
@@ -369,6 +370,14 @@ printerOptsMeta =
             metaPlaceholder = "BOOL",
             metaHelp = "Give the programmer more choice on where to insert blank lines",
             metaDefault = True
+          },
+      poUnicode =
+        PrinterOptsFieldMeta
+          { metaName = "unicode",
+            metaGetField = poUnicode,
+            metaPlaceholder = "UNICODE",
+            metaHelp = printf "Output Unicode syntax (choices: %s)" (showAllValues unicodePreferenceMap),
+            metaDefault = UnicodeNever
           }
     }
 
@@ -449,6 +458,15 @@ inStyleMap =
       ]
    )
 
+unicodePreferenceMap :: BijectiveMap Unicode
+unicodePreferenceMap =
+  $( mkBijectiveMap
+      [ ('UnicodeDetect, "detect"),
+        ('UnicodeAlways, "always"),
+        ('UnicodeNever, "never")
+      ]
+   )
+
 instance PrinterOptsFieldType CommaStyle where
   parseJSON = parseJSONWith commaStyleMap "CommaStyle"
   parseText = parseTextWith commaStyleMap
@@ -478,6 +496,11 @@ instance PrinterOptsFieldType InStyle where
   parseJSON = parseJSONWith inStyleMap "InStyle"
   parseText = parseTextWith inStyleMap
   showText = show . showTextWith inStyleMap
+
+instance PrinterOptsFieldType Unicode where
+  parseJSON = parseJSONWith unicodePreferenceMap "UnicodePreference"
+  parseText = parseTextWith unicodePreferenceMap
+  showText = show . showTextWith unicodePreferenceMap
 
 ----------------------------------------------------------------------------
 -- BijectiveMap helpers
--- a/src/Ormolu/Printer/Combinators.hs
+++ b/src/Ormolu/Printer/Combinators.hs
@@ -38,6 +38,7 @@ module Ormolu.Printer.Combinators
     breakpoint,
     breakpoint',
     getPrinterOpt,
+    whenUnicodeOtherwise,
 
     -- ** Formatting lists
     sep,
@@ -80,8 +81,10 @@ where
 import Control.Monad
 import Data.List (intersperse)
 import Data.Text (Text)
+import GHC.LanguageExtensions.Type
 import GHC.Types.SrcLoc
 import Ormolu.Config
+import Ormolu.Config.Types
 import Ormolu.Printer.Comments
 import Ormolu.Printer.Internal
 import Ormolu.Utils (HasSrcSpan (..))
@@ -164,6 +167,17 @@ breakpoint = vlayout space newline
 -- > breakpoint' = vlayout (return ()) newline
 breakpoint' :: R ()
 breakpoint' = vlayout (return ()) newline
+
+-- | Write the one text or the other depending on whether Unicode is enabled.
+whenUnicodeOtherwise :: Text -> Text -> R ()
+unicodeText `whenUnicodeOtherwise` asciiText = do
+  unicodePrinterOption <- getPrinterOpt poUnicode
+  unicodeExtensionIsEnabled <- isExtensionEnabled UnicodeSyntax
+  txt $ case unicodePrinterOption of
+    UnicodeDetect | unicodeExtensionIsEnabled -> unicodeText
+    UnicodeDetect | otherwise -> asciiText
+    UnicodeAlways -> unicodeText
+    UnicodeNever -> asciiText
 
 ----------------------------------------------------------------------------
 -- Formatting lists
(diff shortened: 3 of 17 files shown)


brandonchinn178:
Thanks! Can you make the code and config a bit less opinionated? A simple bool for the config, and rename the args to whenUnicodeOtherwise...

georgefst:
This is cool, and I'm interested in trying it on some of my smaller personal projects. But I agree with @brandonchinn178 that more neutral language would be preferable, seeing as the current wording doesn't necessarily reflect all maintainers' views on unicode syntax.

kindaro:
Like so?

brandonchinn178:
Perfect, thanks! Can you follow the instructions in CONTRIBUTING.md to update docs + tests for your new option?

src/Ormolu/Config.hs
```diff
@@ -348,6 +349,14 @@ printerOptsMeta =
             metaPlaceholder = "HEIGHT",
             metaHelp = "Number of spaces between top-level declarations",
             metaDefault = 1
+          },
+      poUnicode =
+        PrinterOptsFieldMeta
+          { metaName = "unicode",
+            metaGetField = poUnicode,
+            metaPlaceholder = "BOOL",
+            metaHelp = "Take full advantage of Unicode syntax",
```
dpwiz, line 358: I don't know what to expect from this flag. What's the advantage?.. The missiles would launch in O(1)?

kindaro:
@brandonchinn178   There is no such file in this repository. Can you link me to it? Better even, tell me what you want me to do.

src/Ormolu/Config.hs
kindaro: What do you suggest?

brandonchinn178:
Sorry, DEVELOPER.md

src/Ormolu/Config.hs
georgefst: It's minor, but I agree that the wording is slightly odd. Perhaps simply "Use unicode syntax"? Or "Output unicode syntax".

georgefst:
I think this might actually work best as a ternary option, with the values `always`, `never` and `detect`. Where `detect` will output unicode if and only if the `UnicodeSyntax` extension is enabled, similarly to what we do with `ImportQualifiedPost`. I feel `detect` would be the sensible default.

georgefst:
Is there any particular reason why this doesn't cover [all unicode syntax](https://downloads.haskell.org/~ghc/9.2.1/docs/html/users_guide/exts/unicode_syntax.html) e.g. `forall`/``∀`?

I don't mind if the constructs currently implemented are the only ones you care about! But if so, we should open a follow-up issue tracking the fact that we should implement the rest.

src/Ormolu/Config.hs
kindaro: Let it be _«output Unicode syntax»_ then.

kindaro:
* [ ] Change the `metaHelp` text.
* [ ] Detect the `UnicodeSyntax` extension.
* [ ] Add missing Unicode symbols.

kindaro:
So, as George says, `ImportQualifiedPost` is automatic. And it does not have any command line flags to it. If the extension is enabled, the syntax is used — easy!

Why then have a command line option for `UnicodeSyntax`? Maybe we should hard-wire it in the same way as `ImportQualifiedPost`?

brandonchinn178:
That makes sense to me! Have you checked to see if this is something Ormolu wants? If so, it would be better to make the change upstream instead.

georgefst:
I see little harm in it being configurable. But I'm not that bothered either way. If someone has turned on the extension, it does seem odd for them not to want unicode output. Then again, what if `UnicodeSyntax` is added to `GHC202*`?

And I agree that if we're going to go with the non-configurable version, then it's worth at least mentioning upstream.

brandonchinn178:
> I see little harm in it being configurable.

That's fair. I think this is something Ormolu would be interested in anyway, so I think we should wait and see how that conversation pans out. If they don't want it, we can merge it here. If they do, we could think about making an option to forcibly turn it on.

kindaro:
They do not want it, so we are back here.

I say there should only be two options: _«automatic»_ and _«never»_. If you try to add Unicode syntax to a module where the extension is not enabled, there will be a parse error, so it is pointless to add an option _«always»_. Agreeable?

georgefst:
> If you try to add Unicode syntax to a module where the extension is not enabled, there will be a parse error, so it is pointless to add an option _«always»_.

My concern was that a user's environment might be set up such that Fourmolu doesn't know the extension is active. But in that case, I guess they can just pass the extension to Fourmolu explicitly.

brandonchinn178:
Personally, I dont think we need any option. If someone wants to disable it for a particular module, they can turn off the extension with NoUnicodeSyntax

georgefst:
I have a weak preference towards the ternary option, but I don't want to bikeshed. I'm happy with no option if that's what others prefer. We can always add options later if someone has a use case.

kindaro:
1. It makes sense that `fourmolu` should behave exactly as `ormolu` with some _«compatibility»_ configuration But `ormolu` declined to add this feature. Therefore, we must have at least a setting to switch this feature off.
2. The extension `UnicodeSyntax` can be supplied to `fourmolu` with a specialized option, which will result in this feature being switched on for all modules. Therefore, the option to switch this feature on is redundant.

I shall add options `off` and `auto`. Hopefully soon.

kindaro:
Another argument for giving a flag to switch the extension off: currently, if you enable `ImportQualifiedPost`, `fourmolu` will format all your imports that way — then, if you remove the extension, it could not parse the code anymore! So, there is no way back, even though in principle `fourmolu` can as easily format the imports back.

kindaro:
Hey I have been busy but finally — rebased, added a three way option, and automatic detection of the extension. Ready for review!

src/Ormolu/Config.hs
kindaro: Changes made, resolving.

src/Ormolu/Config.hs
```diff
@@ -402,6 +420,11 @@ instance PrinterOptsFieldType ImportExportStyle where
   parseText = parseTextWith importExportStyleMap
   showText = show . showTextWith importExportStyleMap
 
+instance PrinterOptsFieldType Unicode where
+  parseJSON = parseJSONWith unicodePreferenceMap "CommaStyle"
```
brandonchinn178, line 424:
```suggestion
  parseJSON = parseJSONWith unicodePreferenceMap "UnicodeSyntax"
```

src/Ormolu/Printer/Combinators.hs
```diff
@@ -165,6 +168,17 @@ breakpoint = vlayout space newline
...
 breakpoint' = vlayout (return ()) newline
 
+-- | Write the one text or the other depending on whether Unicode is enabled.
+whenUnicodeOtherwise :: Text -> Text -> R ()
+unicodeText `whenUnicodeOtherwise` asciiText = do
+  unicodePrinterOption <- getPrinterOpt poUnicode
+  unicodeExtension <- isExtensionEnabled UnicodeSyntax
+  txt $ case (unicodePrinterOption, unicodeExtension) of
+    (UnicodeDetect, True) -> unicodeText
+    (UnicodeDetect, False) -> asciiText
```
brandonchinn178, line 178: Nit: if `UnicodeDetect` is the only one needing `unicodeExtension`, you could use guards instead of forcing the `UnicodeAlways`/`UnicodeNever` branches to ignore the second argument

brandonchinn178 (commented):
Looks great, thanks! Please make sure to do everything in this list: https://github.com/fourmolu/fourmolu/blob/main/DEVELOPER.md#adding-a-new-configuration-option

brandonchinn178:
Also, it looks like there are test failures. The biggest one being that `fourmolu.yaml` should be updated to include `unicode: never`

src/Ormolu/Config.hs
kindaro: Oh whoops.

src/Ormolu/Printer/Combinators.hs
kindaro: Yes, let us do guards.

kindaro:
Incidentally from the failing test case I noticed that the linear lollipop is missing, so I should add that one too.

src/Ormolu/Config.hs
kindaro: Fixed.

src/Ormolu/Printer/Combinators.hs
kindaro: Fixed.

brandonchinn178 (commented):
Looks like you still need to do the "Adding a new configuration option" steps in DEVELOPER.md, e.g. updating the README and CHANGELOG and adding tests

kindaro:
Oh, totally forgot about that.

kindaro:
@brandonchinn178   Done!

data/fourmolu/unicode-syntax/output-UnicodeAlways.hs
```diff
@@ -0,0 +1,90 @@
...
+
+data Tree a = Leaf a | Branch (Tree a) (Tree a)
+
+instance (Eq a) ⇒ Eq (Tree a) where
+    Leaf a == Leaf b = a == b
+    (Branch l1 r1) == (Branch l2 r2) = (l1 == l2) && (r1 == r2)
+    _ == _ = False
+
+add1 ∷ Quote m ⇒ Int → m Exp
+add1 x = [|x + 1⟧
```
brandonchinn178, line 71: should the opening quasiquote here be unicode?

tests/Ormolu/Config/PrinterOptsSpec.hs
```diff
@@ -113,6 +113,13 @@ spec =
             show newlines ++ if respectful then " (respectful)" else "",
           testCaseSuffix = \(newlines, respectful) ->
             suffixWith [show newlines, if respectful then "respectful" else ""]
+        },
+      TestGroup
+        { label = "unicode-syntax",
```
brandonchinn178, line 118: what do you think about two test suites, one with `input.hs` with `-XUnicodeSyntax` and one without? alternatively, maybe we could support multiple `input.hs` files (e.g. `input-unicode.hs` and `input-ascii.hs`)

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 ## Unreleased
 
 * Add `emptyConfig`
+* Add `unicode` option.
```
brandonchinn178, line 4: Can you include a link to this PR? Also, can you resolve the merge conflict from https://github.com/fourmolu/fourmolu/pull/223?

brandonchinn178 (commented):
Seems like there are still some test failures

CHANGELOG.md
kindaro: Yes sure.

tests/Ormolu/Config/PrinterOptsSpec.hs
kindaro: Yes sure give me one minute.

data/fourmolu/unicode-syntax/output-UnicodeAlways.hs
kindaro: Yes it should. I shall investigate.

kindaro:
> Seems like there are still some test failures

Yes, I changed some white space in `input.hs` and forgot to refresh the output files because I thought it should not make difference but turns out it does.

It is the small things that will be my undoing.

data/fourmolu/unicode-syntax/output-UnicodeAlways.hs
kindaro: For some reason, when I make the left bracket into Unicode, idempotence breaks. It will take some time to get to the bottom of this.

CHANGELOG.md
brandonchinn178: Sorry, actually I'm changing the way we do CHANGELOG entries (because I'm noticing how many merge conflicts are happening). Please use the new format mentioned here: https://github.com/fourmolu/fourmolu/pull/228

data/fourmolu/unicode-syntax/output-UnicodeAlways.hs
kindaro: The problem is in Ormolu as well — the parser parses `⟦` as `[e|` and then we must print it as `[e|` as well. Issue to Ormolu: <https://github.com/tweag/ormolu/issues/934>.

tests/Ormolu/Config/PrinterOptsSpec.hs
kindaro: @brandonchinn178   On a second thought, I do not think this makes sense.

If we set `unicode` to `always`, the file will get formatted in Unicode, but it will not parse when `UnicodeSyntax` is off. We could check only for `never` and `detect` but I do not see how to do it with the machinery provided by `PrinterOptsSpec.hs` and it does not seem to worth writing a bespoke test case.

Please explain if you have a particular vision — I do not understand what to do about this yet.

kindaro:
I rebased, there are no test failures, `changelog.d` updated, ready to ship.

data/fourmolu/unicode-syntax/output-UnicodeAlways.hs
kindaro: Ormolu fixed their problem. Once it it merged to `fourmolu`, I can tweak the code so that `⟦x⟧` is handled correctly.

brandonchinn178 (approved):

brandonchinn178:
Thanks!

src/Ormolu/Printer/Combinators.hs
```diff
@@ -336,6 +365,71 @@ commaDel' = \case
...
+cparenbar = "⦈" `whenUnicodeOtherwise` "|)"
+
+-- | Print @⟦@ or @[|@ as appropriate.
+openExpQuote = "⟦" `whenUnicodeOtherwise` "[|"
+
+-- | Print @⟧@ or @|]@ as appropriate.
+closeQuote = "⟧" `whenUnicodeOtherwise` "|]"
+
+-- | Print @⊸@ or @%1 ->@ as appropriate.
+lolly = "⊸" `whenUnicodeOtherwise` "%1 ->"
```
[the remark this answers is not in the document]
kindaro: Are you sure? This naming is not my idea. [It is called `lolly` in GHC's source code.][1]

[1]: https://gitlab.haskell.org/ghc/ghc/-/blob/080fffa1015bcc0cff8ab4ad1eeb507fb7a13383/compiler/GHC/Parser/Lexer.x#L811

src/Ormolu/Printer/Combinators.hs
```diff
@@ -165,6 +183,17 @@ breakpoint = vlayout space newline
...
 breakpoint' = vlayout (return ()) newline
 
+-- | Write the one text or the other depending on whether Unicode is enabled.
+whenUnicodeOtherwise :: Text -> Text -> R ()
+unicodeText `whenUnicodeOtherwise` asciiText = do
+  unicodePrinterOption <- getPrinterOpt poUnicode
+  unicodeExtensionIsEnabled <- isExtensionEnabled UnicodeSyntax
+  txt $ case unicodePrinterOption of
+    UnicodeDetect | unicodeExtensionIsEnabled -> unicodeText
+    UnicodeDetect | otherwise -> asciiText
```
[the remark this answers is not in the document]
kindaro: Yes, `if` would be good.

src/Ormolu/Printer/Combinators.hs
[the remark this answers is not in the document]
brandonchinn178: oh interesting. Ok 🤷

Issue #201, opened 2022-06-01
Consolidate import-export-comma-style and diff-friendly-import-export


brandonchinn178:
https://github.com/fourmolu/fourmolu/pull/108#issuecomment-1100916957
https://github.com/fourmolu/fourmolu/pull/169#issuecomment-1141459894

These two options aren't completely independent; `import-export-comma-style=Leading` and `diff-friendly-import-export=True` is probably not a format anyone wants. We might want to instead have a single configuration option `import-export-style: leading, trailing, diff-friendly`

3kyro:
I'm in favor of the single configuration despite the breakage. We could also provide automatic modification of config files with a deprecation notice that will be kept for a couple of versions.

brandonchinn178:
I would weakly prefer not doing any automatic modification of config files; seems like unnecessary complexity. But yes, that all sounds good

3kyro:
great, I'll work on it in the next days.

georgefst:
Closed by #207.

Although I would quite like to see a shell script or something (cabal script?) that could be used to migrate config files. This would set a good precedent.

PR #207, opened 2022-06-19
Consolidate import export comma styles

3kyro:
Consolidate `import-export-comma-style` and `diff-friendly-import-export` into a new option `import-export-style` as per #201

cc @brandonchinn178


src/Ormolu/Printer/Meat/Module.hs
```diff
@@ -23,6 +23,8 @@ import Ormolu.Printer.Meat.Declaration.Warning
 import Ormolu.Printer.Meat.ImportExport
 import Ormolu.Printer.Meat.Pragma
 
+{- HLINT ignore "Use camelCase" -}
```
brandonchinn178, line 26: Can we avoid adding hlint stuff to the repo? We are explicitly not hlint-compatible (see DEVELOPER.md)

src/Ormolu/Config/Types.hs
```diff
@@ -43,3 +45,22 @@ data HaddockPrintStyle
...
+
+data ImportExportStyle
+  = ImportExportLeading
+  | ImportExportTrailing
+  | ImportExportDiffFriendly
+  deriving (Eq, Show, Enum, Bounded)
+
+-- poImportExportStyle helpers.
+
+data DiffFriendly = DiffFriendly
```
brandonchinn178, line 57: this doesnt seem to be used?

src/Ormolu/Config/Types.hs
```diff
@@ -43,3 +45,22 @@ data HaddockPrintStyle
...
+
+isDiffFriendly :: ImportExportStyle -> Bool
+isDiffFriendly = (==) ImportExportDiffFriendly
+
+importExportCommaStyle :: ImportExportStyle -> Maybe CommaStyle
+importExportCommaStyle = \case
+  ImportExportLeading -> Just Leading
+  ImportExportTrailing -> Just Trailing
+  ImportExportDiffFriendly -> Nothing
```
brandonchinn178, line 66: Shouldnt DiffFriendly just map to Trailing? This also only seems to be used in one location, can we just inline there?

src/Ormolu/Printer/Meat/ImportExport.hs
```diff
@@ -209,8 +210,17 @@ parens' topLevelImport m =
...
     trailingParen = if topLevelImport then txt " )" else txt ")"
 
+getCommaStyle :: R CommaStyle
+getCommaStyle = do
+  style <- getPrinterOpt poImportExportStyle
+  pure $ fromMaybe Trailing $ importExportCommaStyle style
+
+getDiffFriendly :: R Bool
+getDiffFriendly =
+  isDiffFriendly <$> getPrinterOpt poImportExportStyle
```
brandonchinn178, line 220: this is only used in one location, can we just inline?

brandonchinn178 (commented):
Looks great! A few comments; can you also make sure to update the "Specifying config" section in the README too?

src/Ormolu/Printer/Meat/Module.hs
3kyro: Done, I've also removed an earlier one I've sneaked in `Meat/ImportExport.hs` during the leading commas PR.

src/Ormolu/Config/Types.hs
3kyro: Removed

src/Ormolu/Config/Types.hs
3kyro: Yes, it simplifies things a bit. Done.

src/Ormolu/Printer/Meat/ImportExport.hs
3kyro: done

src/Ormolu/Printer/Meat/ImportExport.hs
```diff
@@ -192,10 +190,11 @@ parens' topLevelImport m =
       body
       txt ")"
   where
+    getDiffFriendly = isDiffFriendly <$> getPrinterOpt poImportExportStyle
```
brandonchinn178, line 193: Inline this as well? Should be able to get rid of isDiffFriendly altogether

brandonchinn178 (commented):
I think you're still missing the README change I asked about, in the "Specifying configuration" section

src/Ormolu/Printer/Meat/ImportExport.hs
3kyro: ok, inlined at both use sites

3kyro:
> I think you're still missing the README change I asked about, in the "Specifying configuration" section

ah, yes I did :(. I'm not sure why the fourmolu style was mentioning leading commas, but default configuration should be `diff-friendly`.

brandonchinn178 (approved):
Looks great, thank you!

PR #95, opened 2021-06-12
Leading arrows

expipiplus1:
This adds a --leading-arrows option and closes https://github.com/fourmolu/fourmolu/issues/86

~~It's currently on top of my alignment branch for my convenience, I don't
think it should be that hard to extract though~~

I've extracted it from on top of my align work, although it's a more convenient for me to have it on there. I'll be keeping that up to date on this branch (joe-align-leading) https://github.com/expipiplus1/fourmolu/tree/joe-align-leading

CC @dpwiz

Issue #86, opened 2021-04-29
Leading arrows in types


dpwiz:
Basically a revert of the [original PR](https://github.com/tweag/ormolu/pull/392/files), put under an option.

There is another fork that appears to track this: https://github.com/google/ormolu

expipiplus1:
As much as I love leading arrows in types, I do think there's value in having the `::` on the same line as the identifier as it makes grepping (when HLS isn't working) for some definition easier. Is it possible to have an option for this too when you make your PR @dpwiz :wink:

expipiplus1:
I cherry picked the leading arrow change from that fork, it's on top of my alignment branch here https://github.com/expipiplus1/fourmolu/tree/joe-align-leading-arrow

dpwiz:
Cool. Making it optional is where I crashed :sweat_smile:

PR #209, opened 2022-06-21
Support leading arrows

3kyro:
This is the rebasing branch for #95 . ~~Very much work in progress~~
Remaining items:

- [x]  Rebase `Meat/Type`
- [x] Fix/add tests
- [x] Update Changelog
- [x] Update README


3kyro:
> I tried this and get things like:
> 
> ```haskell
> commandCallbacks :: :: [Callback UpdateIssue (Maybe (M.Issue Result)) State]
> ```

Hey, sorry but I'm still in the process of rebasing the original PR, so it's bound to be broken in some way 😃

3kyro:
@expipiplus1 I've also created a PR that targets your original branch :)

expipiplus1:
Thanks, probably best to stick to this PR! Not sure of the value of me merging into my branch at this point haha

arybczak:
Ping. Is this still Wip as the title suggests?

3kyro:
> Thanks, probably best to stick to this PR! Not sure of the value of me merging into my branch at this point haha

@brandonchinn178 I think this PR can be reviewed now. I've added a small set of tests in https://github.com/fourmolu/fourmolu/pull/209/commits/9e72445b3397c7f61a78a31ff273a2ef7d971e9e that should probably be expanded a bit (maybe @expipiplus1 could help :) )

cc @arybczak

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 ## Unreleased
 
 * Add `emptyConfig`
+* Add `leading-arrows` configuration option to allow leading arrows in type signatures
```
brandonchinn178, line 4: Can you add a link to this PR?

README.md
```diff
@@ -52,6 +52,7 @@ We share all bar one of Ormolu's goals:
 | `haddock-style`      | `multi-line`, `single-line` | Whether multiline haddocks should use `{-` or `--` |
 | `newlines-between-decls` | any integer | number of newlines between top-level declarations |
 | `fixities`           | A list of strings for defining fixities; see the "Language extensions, dependencies, and fixities" section below |
+| `leading-arrows`           | `true`, `false`| How to format arrows in type signatures |
```
brandonchinn178, line 55: Nit: please align the first column with the rest of the table

README.md
```diff
@@ -71,6 +72,7 @@ respectful: true
 haddock-style: multi-line
 newlines-between-decls: 1
 fixities: []
+fixities: false
```
brandonchinn178, line 75: `leading-arrows`? Same below

data/fourmolu/leading-arrows/input.hs
```diff
@@ -0,0 +1,74 @@
...
+    forall a b.
+    (Show a, Eq b) => -- foo
+    -- bar
+    a ->
+    b ->
+    Foo 'Int
+  -- | But 'Bar' is also not too bad.
+  Bar ::
+    Int ->
+    Maybe Text ->
```
brandonchinn178, line 38: Can we include haddocks in this test input?

data/fourmolu/leading-arrows/input.hs
```diff
@@ -0,0 +1,74 @@
...
+  -- | So is 'Baz'.
+  Baz ::
+    forall a.
+    a ->
+    Foo 'String
+  (:~>) :: Foo a -> Foo a -> Foo a
+
+instance Eq Int where
+  (==) :: Int -> Int -> Bool
+  (==) _ _ = False
```
brandonchinn178, line 49: what's the point of this?

data/fourmolu/leading-arrows/output-True.hs
```diff
@@ -0,0 +1,75 @@
...
+        :: Int
+        -> Int
+        -> Ordering
+    compare
+        _
+        _ =
+            GT
+
+functionName
+    :: (C1, C2, C3, C4, C5)
```
brandonchinn178, line 63: I think there's value in keeping `::` next to the identifier: https://github.com/fourmolu/fourmolu/issues/86#issuecomment-855669208

Similar to [what happened](https://github.com/fourmolu/fourmolu/issues/201) with `diff-friendly-import-export`, it would be nice if `leading-arrows` didn't change two things: the placement of `::` and the placement of `->`. I think it would be more consistent, style-wise, to do something like
```hs
foo ::
  a
  -> String
  -> Bool

foo2 ::
  Show a =>
  a
  -> String
  -> Bool

foo3 ::
  ( Show a
  , Eq a
  ) =>
  a
  -> String
  -> Bool

foo4 ::
  forall a.
  ( Show a
  , Eq a
  ) =>
  a
  -> String
  -> Bool
```
My reasoning:
* The main part of the type `a -> String -> Bool` is visually separate from the constraints and is formatted the same, whether there are constraints or not
* Keeping the fat arrow with the multiline constraints also avoids a lonely `)` on its own line
* Similarly, constraints are formatted the same whether there's an explicit `forall` or not
* Avoids double indenting when constraints are on multiple lines:
    ```hs
    -- the constraints are all an extra three spaces over (5 spaces total from `foo`)
    foo
      :: ( MonadLogger m
         , MonadIO m
         , MonadCatch m
         , MonadReader MyEnv m
         )

    -- vs. only two spaces from `foo`
    foo ::
      ( MonadLogger m
      , MonadIO m
      , MonadCatch m
      , MonadReader MyEnv m
      )
    ```

Thoughts?

src/Ormolu/Printer/Internal.hs
```diff
@@ -443,6 +444,15 @@ inci = inciByFrac 1
 inciHalf :: R () -> R ()
 inciHalf = inciByFrac 2
 
+-- | Increase indentation by 3.
+inci3 :: R () -> R ()
```
brandonchinn178, line 448: Instead of hardcoding this, maybe export a `inciByExact`? This feels like a magic-number situation

brandonchinn178 (commented):
Thanks for the PR! Still looking through, left some initial thoughts

data/fourmolu/leading-arrows/output-True.hs
expipiplus1: :-1: to this IMO. This design has been in the making for more than a year and it would be a real shame to delay it any further over what seems to me like quite a controversial last-minute suggestion.

While I think that a strong argument can be made for keeping `::` on the same line as the function name (to aid greppability), I think it would be horrible to end up with something like the following:

```haskell
foo ::
  C =>
  D =>
  E =>
  -> f
  -> g
```

This isn't some weird case nobody would ever write, I've observed this pattern in real-world code.

*I definitely wouldn't object to a future PR, adding a `no-::-or-=>` option alongside `true` and `false`*

brandonchinn178: > This design has been in the making for more than a year and it would be a real shame to delay it any further over what seems to me like quite a controversial last-minute suggestion.

Where was it discussed for this year? I have seen zero discussion about this, so from my perspective, this is a big design decision that I'm being asked to merge with zero feedback.

expipiplus1: https://github.com/fourmolu/fourmolu/pull/95

expipiplus1: Sorry the the brusque tone! Thank you for the work you're doing.

brandonchinn178: No worries, just want to make sure we're on the same page. I did see the other PR, and I do recognize it's been stalled for over a year. But I dont see any discussion in that PR (or anywhere) on specifically how to style things. I think it's important to have this discussion later than never.

If your :-1: is truly because of the style, I'd like to hear why. But I dont think we should shoot down changes just because it's been a year in the making.

To respond to your example,

```hs
foo ::
  C =>
  D =>
  E =>
  f
  -> g
```
I think this looks perfectly fine (as fine as it could look with multiple constraint bodies).

But I'd be willing to compromise on the fat arrows, if we could keep the double colon and forall-period on the previous line. IMO this would look nicer, and be more aligned with the option named "leading-arrows"

georgefst: I have to agree with @expipiplus1. IME, when people talk about "leading arrows" in Haskell, they have a specific design in mind which includes leading `::`, `=>` etc. (I personally hate it, but that's neither here nor there)

Perhaps we could get people who have expressed an interest in using this feature (@ocharles, @dpwiz, @Disco-Dave, @arybczak, @parsonsmatt, @kalhauge, @Philonous, @eltix, @tschuchortdev) to cast an eye over the test suite changes and check that output is in line with their expectations.

brandonchinn178: Sure, that makes sense. I would certainly use a `leading-arrows: mixed` / `leading-arrows: only-args` option myself (i.e. keep `::`, forall-periods, and fat arrows as trailing), but I can do that as a followup

@3kyro does/should this apply for function types in record fields? Should we add a test for that as well?

src/Ormolu/Printer/Combinators.hs
```diff
@@ -364,3 +369,24 @@ placeHanging placement m =
     Normal -> do
       breakpoint
       inci m
+
+----------------------------------------------------------------------------
+-- Arrow style
+
+-- | Ouput @space >> txt "::" >> x@ when we are printing with trailing arrows
```
brandonchinn178, line 376:
```suggestion
-- | Output @space >> txt "::" >> x@ when we are printing with trailing arrows
```
same below

src/Ormolu/Printer/Combinators.hs
```diff
@@ -364,3 +369,24 @@ placeHanging placement m =
...
+trailingArrowType :: R () -> R ()
+trailingArrowType x = do
```
brandonchinn178, line 378: Seems like `x` is always `pure ()`? Can get rid of it?

src/Ormolu/Printer/Meat/Type.hs
```diff
@@ -23,175 +23,229 @@ module Ormolu.Printer.Meat.Type
...
+p_hsTypePostDoc t = p_hsType' AfterNothing (hasDocStrings t) CaretStyle t
 
 -- | How to render Haddocks associated with a type.
 data TypeDocStyle
   = PipeStyle
   | CaretStyle
 
-p_hsType' :: Bool -> TypeDocStyle -> HsType GhcPs -> R ()
-p_hsType' multilineArgs docStyle = \case
-  HsForAllTy _ tele t -> do
```
brandonchinn178, line 53: the indentation change is gonna cause ugly merge conflicts with ormolu. can we break out a separate `p_hsType2` (named better) that does the `p_leading >> p_hsType1` logic?

data/fourmolu/leading-arrows/output-True.hs
arybczak: For the record, what @brandonchinn178 are then no longer true leading arrows because of `=>` ;)

So I'd like to have the current design for `true`, the `mixed` option can be added later.

kalhauge: I think the test-suite as present, is what you would expect with turning `leading-arrows: true`, great work!  However, I'm with @brandonchinn178 , that the solution that I ultimately want is something like `only-args`, potentially with the first argument indented:

```haskell
fun ::
   forall a b. (C1, C2) => 
      a -- ^ first argument
   -> c -- ^ second argument
```

Which might be expanded to:

```haskell
fun ::
   forall a b. 
    ( C1
    , C2
    ) => 
      a -- ^ first argument
   -> c -- ^ second argument
```

3kyro: @brandonchinn178 @georgefst @expipiplus1 Seems like a way forward could be to change the config type from a boolean to a custom `function-arrows:leading/trailing` which can be expanded in the future but keeps  the current implementation as `leading`. Would that be OK with you? I'm currently traveling but can have a look along with the other comments early next week.

> @3kyro does/should this apply for function types in record fields? Should we add a test for that as well?

Seems like function types in record field still need some work (see update test suite)

brandonchinn178: Sounds good with me!

expipiplus1: That sounds like an excellent idea, @3kyro :)

dpwiz: Leaving `::` or `=>` dangling will diverge from the style `align type stuff on the left`. I don't think I've ever seen such code in the wild. Personally, I would be very surprised (and, honestly, annoyed) if applying a "leading-arrows" config to my codebase would produce a diffsplosion introducing the shuffle.

`function-arrows: leading/trailing` is nice, though.

src/Ormolu/Printer/Meat/Type.hs
```diff
@@ -23,175 +23,229 @@ module Ormolu.Printer.Meat.Type
...
+  getPrinterOpt poLeadingArrows >>= \case
+    False -> pure ()
+    True -> p_after multiline a
+
 p_hsType :: HsType GhcPs -> R ()
-p_hsType t = p_hsType' (hasDocStrings t) PipeStyle t
+p_hsType = p_hsTypeAfter' AfterNothing
+
+-- | Like 'p_hsType' but indent properly following a forall
+p_hsTypeAfter' :: After -> HsType GhcPs -> R ()
```
brandonchinn178, line 72: It seems like this is only used in `p_hsType`. Can we inline this?

src/Ormolu/Printer/Meat/Type.hs
```diff
@@ -23,175 +23,229 @@ module Ormolu.Printer.Meat.Type
...
+import Ormolu.Printer.Internal (enterLayout)
 import Ormolu.Printer.Meat.Common
 import {-# SOURCE #-} Ormolu.Printer.Meat.Declaration.OpTree (p_tyOpTree, tyOpTree)
 import {-# SOURCE #-} Ormolu.Printer.Meat.Declaration.Value (p_hsSplice, p_stringLit)
 import Ormolu.Printer.Operators
 import Ormolu.Utils
 
+-- What precedes this type on the same line.
+-- Only used for leading arrows
+data After
```
brandonchinn178, line 45: What do you think about putting this into the `SC` context (as `prevTypeCtx :: Maybe TypeCtx`, where `AfterNothing` => `Just TypeCtxStart`, `AfterForall` => `Just TypeCtxForall`, etc.) and provide helpers
```hs
-- for p_hsType and p_hsTypePostDoc
withPrevTypeCtx :: R () -> R ()
withPrevTypeCtx m = do
  R $ modify (\sc -> sc{prevTypeCtx = Just TypeCtxStart})
  m
  R $ modify (\sc -> sc{prevTypeCtx = Nothing})

-- only for things inside withPrevTypeCtx
getPrevTypeCtx :: R TypeCtx
getPrevTypeCtx = R (gets (resolve . prevTypeCtx))
  where
    -- or if we dont like partial functions, default to `TypeCtxStart`
    resolve = fromMaybe (error "unexpectedly called getPrevTypeCtx")

-- only for things inside withPrevTypeCtx
setPrevTypeCtx :: TypeCtx -> R ()
```
I think this would reduce the amount of churn in this code (e.g. changing number of arguments to functions, needing to return values from `p_forallBndrs`, etc.) which will make merging upstream changes easier

src/Ormolu/Printer/Meat/Type.hs
brandonchinn178: For reference, I had to do something similar [here](https://github.com/fourmolu/fourmolu/pull/135/files#diff-342756c066801f378dd98d3183b6643ee351ecc9e6083e19f059e35a523ea7e2R144-R146).

Possibly something like
```hs
-- analogous to ormolu's p_hsType'
p_hsType' :: Bool -> TypeDocStyle -> HsType GhcPs -> R ()
p_hsType' multilineArgs docStyle ty = p_leadingAfter multilineArgs >> p_hsTypeImpl multilineArgs docStyle ty

-- original p_hsType' implementation
p_hsTypeImpl :: Bool -> TypeDocStyle -> HsType GhcPs -> R ()
```

data/fourmolu/leading-arrows/output-True.hs
ocharles: The lack of `::` on the same line as the symbol being defined has actually made me reconsider whether I want this. I appreciate all the thoughts about how to get that, but I will probably not end up using leading-arrows after all. That said, one more suggestion from a colleague at work was to do things like:

```haskell
foo :: ()
  => X
  -> Y
  -> Z
```

which I thought was quite clever. You basically always have a `=>` with this style.

kalhauge: I don't know if this is of topic, of this pull request, but I really like @ocharles suggestion, however it has a drawback, namely that it will introduce syntax to a program. I'm not sure if this is against the goal of the project. You could claim that the user just have to all it themselves, they want even prettier aligning of arrows. This means that the following are both valid formatted functions:

```haskell
uglyFun :: 
     a 
  -> c

prettyFun :: ()
  => a 
  -> c
```

The solution do however require that 'preamble' (foralls and constraints) and args are treated differently. namely that the preamble can be on one line while the args are on multiple lines and visa versa, essentially that all of these examples are valid results depending on the indentation of the user:

```haskell
fun :: forall a b. ( C1, C2 ) => a -> c

fun :: forall a b. ( C1, C2 ) 
  => a -> c

fun :: forall a b. ( C1, C2 ) 
  => a 
  -> c


fun :: forall a b. 
  ( C1, C2 ) 
  => a -> c

fun :: forall a b. 
  ( C1
  , C2
  ) 
  => a -> c


fun :: forall a b. 
  ( C1, C2 ) 
  => a 
  -> c

fun :: forall a b. 
  ( C1
  , C2
  ) 
  => a 
  -> c

fun :: 
  forall a b. 
  ( C1
  , C2
  ) 
  => a 
  -> c
```

README.md
3kyro: done

CHANGELOG.md
3kyro: OK, new entry to `changelog.d` added

README.md
3kyro: fixed

data/fourmolu/leading-arrows/input.hs
3kyro: Added haddocks

data/fourmolu/leading-arrows/input.hs
3kyro: I wanted to have a single line type signature as well. I've just copied the first one I've seen tbh, so if it's confusing I can replace with something simpler?

src/Ormolu/Printer/Combinators.hs
3kyro: done

src/Ormolu/Printer/Combinators.hs
3kyro: done

src/Ormolu/Printer/Internal.hs
3kyro: Replace with `inciByExact`

src/Ormolu/Printer/Meat/Type.hs
3kyro: Thanks, I've replaced `After` with `PrevTypeCtx`. However  I've not used a `Maybe` as it seems simpler to always default to `TypeCtxStart` and there is a similar precedent in the `scRequestedDelimiter` field which also simply defaults to `VeryBeginning`.

src/Ormolu/Printer/Meat/Type.hs
3kyro: done

src/Ormolu/Printer/Meat/Type.hs
3kyro: should be addressed with the `PrevTypeCtx` change

3kyro:
@brandonchinn178 I think I've addressed your initial comments, keeping the existing behavior for `function-arrows:leading` as discussed. Tomorrow I'll format our sufficiently large work codebase to have a look for edge cases as well.

data/fourmolu/function-arrows/input.hs
```diff
@@ -0,0 +1,83 @@
...
+  -- | So is 'Baz'.
+  Baz ::
+    forall a.
+    a ->
+    Foo 'String
+  (:~>) :: Foo a -> Foo a -> Foo a
+
+instance Eq Int where
+  (==) :: Int -> Int -> Bool
+  (==) _ _ = False
```
brandonchinn178, line 51: Continuing conversation from https://github.com/fourmolu/fourmolu/pull/209#discussion_r944082764

Can the "test where type is on one line" be a normal function instead of a class instance? Or possibly one of both? And also add a comment saying "testing one-line type"?

src/Ormolu/Printer/Meat/Type.hs
```diff
@@ -51,21 +78,32 @@ data TypeDocStyle
...
-    case tele of
-      HsForAllInvis _ bndrs -> p_forallBndrs ForAllInvis p_hsTyVarBndr bndrs
-      HsForAllVis _ bndrs -> p_forallBndrs ForAllVis p_hsTyVarBndr bndrs
+    a <- case tele of
+      HsForAllInvis _ bndrs -> p_forallBndrs' ForAllInvis p_hsTyVarBndr bndrs
+      HsForAllVis _ bndrs -> p_forallBndrs' ForAllVis p_hsTyVarBndr bndrs
+    a' <-
+      getPrinterOpt poFunctionArrows >>= \case
+        LeadingArrows | multilineArgs -> pure a
+        _ -> p_after False a $> TypeCtxStart
```
brandonchinn178, line 87: Instead of threading the `PrevTypeCtx` everywhere, why not have `p_forallBndrs'` call a new `setPrevTypeCtx` function and `p_after` call `getPrevTypeCtx`?

src/Ormolu/Printer/Meat/Declaration/Signature.hs
```diff
@@ -225,7 +226,7 @@ p_standaloneKindSig (StandaloneKindSig _ name sigTy) = do
   inci $ do
     space
     p_rdrName name
-    space
-    txt "::"
+    trailingArrowType -- TODO, the space in tAT wasn't here before
```
brandonchinn178, line 229: what does this TODO mean?

src/Ormolu/Printer/Meat/Type.hs
```diff
@@ -23,22 +23,49 @@ module Ormolu.Printer.Meat.Type
...
+-- | like p_after but only when we're rendering leading arrows
+p_leadingAfter :: Bool -> PrevTypeCtx -> R ()
+p_leadingAfter multiline a = do
+  getPrinterOpt poFunctionArrows >>= \case
+    TrailingArrows -> pure ()
+    LeadingArrows -> p_after multiline a
+
+-- | Like 'p_hsType' but indent properly following a forall
 p_hsType :: HsType GhcPs -> R ()
-p_hsType t = p_hsType' (hasDocStrings t) PipeStyle t
```
brandonchinn178, line 41: how did it work without `layout == Multiline` here?

src/Ormolu/Printer/Meat/Type.hs
```diff
@@ -23,22 +23,49 @@ module Ormolu.Printer.Meat.Type
...
+p_after :: Bool -> PrevTypeCtx -> R ()
+p_after multilineArgs = \case
+  TypeCtxStart -> pure ()
+  TypeCtxForall | multilineArgs -> txt " ." >> space
+  TypeCtxForall -> txt "." >> space
+  TypeCtxContext -> txt "=>" >> space
+  TypeCtxArgument -> txt "->" >> space
+
+-- | like p_after but only when we're rendering leading arrows
+p_leadingAfter :: Bool -> PrevTypeCtx -> R ()
```
brandonchinn178, line 54: seems like we can inline this?

src/Ormolu/Printer/Meat/Type.hs
```diff
@@ -203,7 +250,9 @@ p_hsType' multilineArgs docStyle = \case
       if multilineArgs
         then newline
         else breakpoint
-    p_hsTypeR = p_hsType' multilineArgs docStyle
+    p_hsTypeR prevTypeCtx m = withPrevTypeCtx prevTypeCtx $ do
```
brandonchinn178, line 253: oh if this is the only place `withPrevTypeCtx` is used, let's just inline it here

brandonchinn178 (commented):
Thanks for getting through all the comments. Just a few more. The trickiest part of this PR is making sure merging Ormolu stuff back isn't going to be a pain. I think if we can resolve the last few comments and utilize `scPrevTypeCtx` more instead of returning + passing `PrevTypeCtx`, we'll be able to reduce the diff even more.

src/Ormolu/Printer/Meat/Declaration/Signature.hs
3kyro: I'm not sure, maybe @expipiplus1 can help? Seems to not be relevant anymore though so I've removed it.

src/Ormolu/Printer/Meat/Type.hs
3kyro: Inlinening it here would require exporting the `R` constructor from `Ormolu.Printer.Internal` though. Maybe we can leave it in the `Internal` module like all the other functions that manipulate the underlying monads of R?

data/fourmolu/function-arrows/input.hs
3kyro: Added `singleLineFun`

src/Ormolu/Printer/Meat/Type.hs
3kyro: Actually implementing `setPrevTypeCtx` makes `withPrevTypeCtx` obsolete

src/Ormolu/Printer/Meat/Type.hs
3kyro: I believe it's needed now for splitting `foralls` and contraints which can't be handled by `interArgBreak` as before.

src/Ormolu/Printer/Meat/Type.hs
3kyro: added `setPrvTypeCtx`

src/Ormolu/Printer/Meat/Declaration/Signature.hs
expipiplus1: honestly no idea, it definitely looks as though there was a space there before... Makes sense to remove it, I guess it was something stale from dev

brandonchinn178:
Looks great! Is this ready to merge, or are you still testing?

3kyro:
> Looks great! Is this ready to merge, or are you still testing?

I think it can be merged. It would be best to squash the commits though as they are a  bit messy due to the rebase.

brandonchinn178 (approved):

3kyro:
@brandonchinn178 Do you have an estimate on when we can expect a release? We're quite eager to switch to fourmolu after this 🙂

brandonchinn178:
I'll make a release within a week. Thanks for the hard work!

expipiplus1:
Thanks @brandonchinn178 and @3kyro!

PR #214, opened 2022-07-16
Resolve directories as inputs (Fixes #213)

### LINKED ISSUE #213: CLI: Support specifying input folders in addition to just input files
It would be nice if fourmolu could take folders as well as files as an input and would process all haskell files in that folder (recursively)
so instead of 
```
fourmolu -i $(find src -name '*.hs')
```
I'd be able to do just 
```
fourmolu -i src
```
This is consistent what formatters in other languages support and would be a nice quality of life improvement.

matobet:


app/Main.hs
```diff
@@ -61,24 +62,42 @@ main = do
           optSourceType
           cfg
 
+      getHaskellFiles input = do
```
brandonchinn178, line 65: None of these helpers use the local scope, right? Can we pull these out to the top level in a "Helpers" section?

app/Main.hs
```diff
@@ -61,24 +62,42 @@ main = do
...
+            if isDir
+              then listDirectoryRecursive child
+              else pure [child]
+
+      selectFailure = \case
+            ExitSuccess -> Nothing
+            ExitFailure n -> Just n
+
+      formatInputs inputs = do
+        files <- sort . nubOrd . join <$> mapM getHaskellFiles inputs
```
brandonchinn178, line 85: Nits:
1. Why not concat instead of join?
2. Instead of `sort . nubOrd`, why not just `Set.toAscList . Set.fromList`? Might as well sort and nub at the same time

brandonchinn178 (commented):
Great work, thanks! Can you add an entry to the CHANGELOG too?

brandonchinn178 (approved):

app/Main.hs
georgefst: Idk, if they're not useful outside of this scope, then I find it cleaner to keep them local anyway.

app/Main.hs
matobet: Done :+1:

app/Main.hs
```diff
@@ -61,24 +61,42 @@ main = do
...
+            if isDir
+              then listDirectoryRecursive child
+              else pure [child]
+
+      selectFailure = \case
+            ExitSuccess -> Nothing
+            ExitFailure n -> Just n
+
+      formatInputs inputs = do
+        files <- Set.toList . Set.fromList . concat <$> mapM getHaskellFiles inputs
```
brandonchinn178, line 84: Sorry, one last nit: I was intentional about saying `toAscList`, since `toList` doesnt have the semantic guarantee that the result will be ordered. Is there a reason to not use that?

I'll merge and make a new release after this last point, promise!

matobet: no worries, fixed :-) :+1:

brandonchinn178:
@matobet oh sorry, I just realized I didn't mention this earlier, but you have a CI failure. Can you fix?

matobet:
@brandonchinn178 sorry missed that among all the green checks :) Now the CI should hopefully pass :+1:

PR #222, opened 2022-08-06
CabalInfo: handle case where source dir is '.'

zyla:
If source dir in Cabal file is `.`, it can happen that `getExtensionAndDepsMap` returns a filename key like `path-to-cabalfile/./Some/Module.hs`, while searching for `path-to-cabalfile/Some/Module.hs`. In this case fourmolu fails to find the appropriate extensions and dependencies for that source file.

This PR fixes the problem by normalizing paths in `getExtensionAndDepsMap`, so that the lookup by source file name (which is also normalize) suceeds.

Also added a test case which reproduces the problem.

### DIFF
--- a/data/cabal-test/cabal-test.cabal
+++ b/data/cabal-test/cabal-test.cabal
@@ -1,0 +1,9 @@
+cabal-version:      2.4
+name:               cabal-test
+version:            0.0.1.0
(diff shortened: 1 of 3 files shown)


brandonchinn178:
This was just fixed in the new 0.8 release :)

PR #229, opened 2022-08-14
Add let-style configuration

### LINKED ISSUE #202: Configuration for styling let
**Describe the option you'd like to add**
* name: `let-style`
    * `inline`: first definition is on the same line as `let`; subsequent definitions are aligned to this definition. If there's an `in` clause, `in` is aligned with start of `let` column, `in` contents is aligned with `let` definitions (i.e. 2 spaces after `in`)
    * `inline-ormolu`: Same as `inline`, except if there's an `in` clause, `in` is aligned with end of `let` column (i.e. 1 space before and after `in`)
    * `auto-newline`: `inline` when there's only one `let` binding, otherwise `newline`
    * `newline`: first definition is on the line after `let`, indented once, with subsequent definitions aligned to this definition. If there's an `in` clause, `in` is aligned with start of `let` column, and the `in` contents start on the next line, indented once
* Ormolu configured with `inline-ormolu`
* I think default behavior should be `auto-newline`, but open to suggestions.

**Show a code example exemplifying the different option settings**
```hs
-- inline
-- note that the space before the `in` is moved to after, not just removed
let a = 1
    b = 2
in  a + b

-- inline-ormolu
let a = 1
    b = 2
 in a + b

-- auto-newline (indent=2)
let a = 1
in  a + 2

-- newline (indent=2)
let
  a = 1
in
  a + 2

-- auto-newline + newline (indent=2)
let
  a = 1
  b = 2
in
  a + b

-- auto-newline + newline (indent=4)
let
    a = 1
    b = 2
in
    a + b
```
Notes:
* With `newline`, we won't support aligning `in` with the end of `let`; it just doesn't really look good. Originally, I wrote this issue where there were multiple independent toggles, but pretty much nothing else looked good. Happy to hear other suggestions as well.
* If someone adds the `in` clause in a do-block, we'll change `"in "` to `" in"` for everything, since keeping it on the same level as `let` is invalid `do`-syntax

brandonchinn178:
Resolves #202 
Supercedes #77 

Still definitely some kinks that could be worked out, especially around empty let blocks (ref: https://github.com/tweag/ormolu/issues/917), let-blocks with multiple bindings on one line (e.g. `let a = 1; b = 2`), etc. But I think this is a good initial implementation.

Examples of each style can be found in the [test directory](https://github.com/fourmolu/fourmolu/tree/let-style/data/fourmolu/let-style). As mentioned in the issue (and in the changelog entry), I made `auto-newline` the default, which changes default formatting. IMO `auto-newline` has the best compromise between Fourmolu's goal of being diff-friendly and compactness. Happy to hear suggestions for another default (I don't think we should use `inline-ormolu`, though, because people are usually confused why `in` starts off-indentation when initially using Fourmolu, e.g. https://github.com/fourmolu/fourmolu/issues/152).


georgefst:
> I don't think we should use inline-ormolu, though, because people are usually confused

I'm gonna go out on a limb and guess that there are just as many people who like the current style, and haven't had reason to complain until now because it's the default (full disclosure, I'm one of them).

More generally, I think we should start being pretty conservative about making changes to default formatting. There are a lot of people using Fourmolu these days and these small changes create a lot of churn. Anecdotally, I know my colleagues would be grateful to hear this.

I am still happy to change defaults for things that are pretty inarguably currently far from ideal e.g. some of the ways in which commented-out code gets moved around, or haddocks with data constructor declarations.

README.md
```diff
@@ -46,6 +46,7 @@ We share all bar one of Ormolu's goals:
 | `indentation`        | any integer   | Number of spaces to use as indentation |
 | `comma-style`        | `leading`, `trailing` | Where to put the comma in lists, tuples, etc. |
 | `import-export-style` | `leading`, `trailing`, `diff-friendly` | How to format multiline import/export lists (`diff-friendly` lists have trailing commas but keep the open-parentheses on the same line as the `import` line) |
+| `let-style`          | `inline`, `inline-ormolu`, `newline`, `auto-newline` | How to style `let` blocks, with `auto-newline` using `inline` for `let` bindings with exactly one binding or `newline` otherwise. See the [test files](https://github.com/fourmolu/fourmolu/tree/main/data/fourmolu/let-style/) for more details. |
```
georgefst, line 49: These options don't quite feel right to me:
- For an "auto" mode, the more conventional Ormolu/Fourmolu thing to do would be to depend on whether the user chooses to insert a newline, rather than the number of bindings. This would also allow us to change the default without reformatting existing code.
- I think we should treat the placement of `in` (Ormolu-style space before vs. alignment of start with `let`) as an orthogonal concern.

brandonchinn178: I was trying to provide a nice, single knob for this option, with my opinion that a shifted `in` with anything other than `inline` doesn't look great. But if people truly like the shifted in (I was always of the mindset: "it'd be nice if it weren't shifted, but I'm _okay_ with it") then I can break that out into a separate option.

I can make it `let-style` (default=`auto`, `inline`, `newline`, `mixed`, with `auto` being as you described and `mixed` as the current `auto-newline`) and `shifted-in` (default=`false`, `true`)

brandonchinn178: Updated: `shifted-in` has been broken out (name suggestions welcome) and `let-style` has been updated with a new `auto` option (which is now the default)

georgefst: Nice. As for name suggestions, maybe just `in-style`?

brandonchinn178: Would that be `in-style: left-align | right-align`?

georgefst: Yes, something like that.

georgefst (approved):

PR #242, opened 2022-10-21
GHC 9.4 support

9999years:
Based on @amesgen's branch here:
https://github.com/tweag/ormolu/pull/885
https://github.com/tweag/ormolu/tree/amesgen/ghc-9.4
https://github.com/tweag/ormolu/commit/0a3da23e9b31e2d45b06d3d2d09a3254acad637e


parsonsmatt:
Looks like we'll need to drop GHC 8.10 support in order to support 9.4.2 - should be fine to delete that part of CI and raise our own lower bound on `base`.

scripts/GetBuildInfo.hs
```diff
@@ -4,7 +4,7 @@
 import Data.Char (toLower)
 import Data.List (intercalate)
 import Distribution.Package (packageVersion)
-import Distribution.PackageDescription.Parsec (readGenericPackageDescription)
+import Distribution.Simple.PackageDescription (readGenericPackageDescription)
```
parsonsmatt, line 7: ahhh this will probably fail when built with earlier `Cabal` - maybe,

```
#if MIN_VERSION_Cabal(3,8,0)
import Distribution.Simple.PackageDescription (readGenericPackageDescription)
#else
import Distribution.PackageDescription.Parsec (readGenericPackageDescription)
#endif
```

(not sure if this is built by all GHC or just most recent)

georgefst:
Thanks for this! We _usually_ wait for Ormolu to merge changes, and periodically merge Ormolu's master branch in to Fourmolu. Often this is when they make a release.

But even if we stick to that process, I'm sure your work here is useful and can be reused.

brandonchinn178:
Thanks! Like @georgefst said, we usually wait for Ormolu to merge. While git should hopefully merge it correctly, if it doesnt, the conflicts can get ugly. Is there any particular reason to merge this into Fourmolu immediately, or is it ok to wait?

Another reason is that it'd be nice to wait until we can actually test with 9.4 in CI.

parsonsmatt:
We use upstream PRs at mercury to track changes in our internal code - we're upgrading to GHC 9.4.2 now, and we need fourmolu to work on it.

9999years:
> While git should hopefully merge it correctly, if it doesnt, the conflicts can get ugly.

There were a fair amount of conflicts getting this PR ported to fourmolu's codebase, but it didn't take more than a few hours to resolve them.

9999years:
[The upstream PR has been merged](https://github.com/tweag/ormolu/pull/885) but not released yet. Do we want to wait for a release to be cut before landing this?

9999years:
We're also integrating these changes into `haskell-language-server` to enable support for `fourmolu` there: https://github.com/haskell/haskell-language-server/pull/3317

brandonchinn178:
For the reasons I mentioned in my last comment, merging this will cause annoying conflicts when merging Ormolu. Hopefully Ormolu makes a release soon and we'll be able to merge it in soon after.

If this happens soon enough, we can just close this PR, as it'll be incorporated in the PR merging Ormolu

brandonchinn178:
Merged in #259. Trying to get #257 merged before making a new release. Will do my best to make a new release shortly

Issue #167, opened 2022-04-13
Put 'where' on same line as ')' in a multiline module export list


tysonzero:
Similar to how leading comma lists have the parenthesis lead as well, it'd be nice to allow this for imports and exports:

```
module Foo
    ( bar
    , baz
    , qux
    ) where

import Foo
    ( bar
    , baz
    , qux
    )
```

brandonchinn178:
isn't this configurable via `diff-friendly-import-export: false`?

tysonzero:
Oh that is closer than I thought:

```
module Foo
    ( bar
    , baz
    , qux
    )
where

import Foo
    ( bar
    , baz
    , qux
    )
```

Only differing thing is the `where` positioning.

brandonchinn178:
Is the `where` positioning of high value to you? While Fourmolu aims to be a configurable formatter, it will never aim to be a maximally configurable formatter. At some point, Fourmolu will be making some decisions that you (or I) might not agree with, but that's what you get with a formatter. IMO, where the `where` goes is a rather minor detail that isn't worth adding a configuration option for.

tysonzero:
Fairly important yeah. Avoiding constructs that I think look weird is like 90% of why I want to put a formatter in our CI process. `where` on the same column as the parent construct is normally a syntax error, so it just feels weird seeing it there even though it's legal in this case.

Yeah it's definitely looking quite difficult to find a true "formatter" that meets my needs. What I really want is closer to a "formatting linter" that doesn't necessarily know how to automatically fix formatting violations, but just spits them to stdout and throws an error for CI. That way it can presumably be significantly more configurable, like `hlint` is.

brandonchinn178:
Oh, sorry, I wasn't aware that that output is what Fourmolu (and Ormolu) does currently. Yes, I do wonder why Ormolu does it that way. I opened an issue asking upstream: https://github.com/tweag/ormolu/issues/876; I suspect that Ormolu will not want to change it, in which case, I think it makes sense to add a configuration option for this.

In other words:
1. If Ormolu closes that issue, this issue should be to put `where` on the previous line
2. If Ormolu fixes it, I'll just close this issue (I can't imagine anyone wanting to configure `where` to be on the next line)

I'll rename this issue in anticipation of (1)

> What I really want is closer to a "formatting linter" that doesn't necessarily know how to automatically fix formatting violations, but just spits them to stdout and throws an error for CI.

[...]

PR #244, opened 2022-10-28
Improve formatting of export list with module warning

brandonchinn178:
Trying to align closer to what Ormolu does here, which should also help with #167 

Blocks #245


georgefst:
Looks good, but I'm struggling to work out why the commits here are split up as they are. Presumably commit 3 is the implementation that causes the output changes in commit 2? So why aren't they squashed together?

brandonchinn178:
ah I was doing TDD, so I fixed the output first to get a test failure, then I just had to experiment in the third commit to make it pass. I can squash them now if you want

georgefst:
Fair enough. It's not my preferred approach but I don't really care.

georgefst (approved):

PR #245, opened 2022-10-29
Respect "respectful" for where keyword in module header

brandonchinn178:
Resolves https://github.com/fourmolu/fourmolu/issues/167
Blocked by #244

### DIFF
--- a/src/Ormolu/Printer/Meat/Module.hs
+++ b/src/Ormolu/Printer/Meat/Module.hs
@@ -12,6 +12,7 @@ import Control.Monad
 import GHC.Hs hiding (comment)
 import GHC.Types.SrcLoc
 import GHC.Unit.Module.Name
+import GHC.Utils.Outputable (ppr, showSDocUnsafe)
 import Ormolu.Config
 import Ormolu.Imports (normalizeImports)
 import Ormolu.Parser.CommentStream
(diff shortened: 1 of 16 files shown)


src/Ormolu/Printer/Meat/Module.hs
```diff
@@ -86,3 +92,14 @@ p_hsModuleHeader HsModule {..} moduleName = do
   breakpointBeforeWhere
   txt "where"
   newline
+  where
+    (moduleKeyword, whereKeyword) =
+      case am_main (anns hsmodAnn) of
+        -- [AnnModule, AnnWhere] or [AnnSignature, AnnWhere]
+        [AddEpAnn _ moduleLoc, AddEpAnn AnnWhere whereLoc] ->
+          (epaLocationRealSrcSpan moduleLoc, epaLocationRealSrcSpan whereLoc)
+        anns -> error $ "Module had unexpected annotations: " ++ showSDocUnsafe (ppr anns)
```
georgefst, line 101: Maybe it's just too long since I've contributed any significant code, but do we generally use `error`? I assumed the main monad stack had some facility for raising pure errors.

georgefst (approved):

src/Ormolu/Printer/Meat/Module.hs
brandonchinn178: For pure errors? Not that I can tell. There's an existing use of `error` already, at least: https://github.com/fourmolu/fourmolu/blob/db6755ae5c0372259ad9aae33a590460a65db0b8/src/Ormolu.hs#L109

PR #246, opened 2022-10-31
Fix README.md examples

clintonmead:
Ensure examples given in README.md don't fail due to hitting command line length limits

### DIFF
--- a/README.md
+++ b/README.md
@@ -135,13 +135,13 @@ $ fourmolu -i Module.hs
 Use `find` to format a tree recursively:
 
 ```console
-$ fourmolu -i $(find . -name '*.hs')
+$ find . -name '*.hs' -exec fourmolu -i {} +
 ```
 
 Or find all files in a project with `git ls-files`:
 
 ```console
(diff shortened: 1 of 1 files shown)


brandonchinn178:
Thanks! Do you mind porting the change to Ormolu? We try to avoid making changes that would conflict with changed Ormolu makes, since we merge upstream code regularly

brandonchinn178:
Actually, I'm just going to close this because we already support specifying directories, which is also mentioned in the README. If someone runs into issues, they should be able to see from the README that there's an alternative

georgefst:
> we already support specifying directories, which is also mentioned in the README

If anything, there's an argument for removing the `find` examples. They make sense for Ormolu, but they're pretty redundant here. They _could_ be useful for when one wants to generalise them, finding only certain files. But we can't document every use case, and too much information can be confusing. Most of our users can figure out how to pipe output between commands. Plus as this PR shows, the examples have issues anyway.

https://github.com/fourmolu/fourmolu/pull/263

PR #247, opened 2022-11-01
Fixes for function arrows

brandonchinn178:
Found some edge cases for function arrows, and was able to simplify the code without the `PrevTypeCtx` machinery.

Some of the changes I made seem to still pass tests (e.g. moving things in/out of a `located` block). If we find a regression or issue later, we can fix it later with an added test

### DIFF
--- a/changelog.d/function-arrows-fixes.md
+++ b/changelog.d/function-arrows-fixes.md
@@ -1,0 +1,1 @@
+Fix some edge cases with function arrows ([#247](https://github.com/fourmolu/fourmolu/pull/247))
--- a/data/examples/declaration/data/gadt/multiline-four-out.hs
+++ b/data/examples/declaration/data/gadt/multiline-four-out.hs
@@ -14,12 +14,8 @@ data Foo a where
         Foo 'Int
     -- | But 'Bar' is also not too bad.
     Bar ::
-        Int ->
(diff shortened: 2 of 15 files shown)


brandonchinn178:
cc @3kyro

3kyro:
@brandonchinn178 There seems to be a regression on multi line arguments with interleaved comments. You can test the following:

````haskell
withComments
  -- Comment
  :: Int
  -- Comment
  -> Int
  -- Comment
  -> Int
```` 
which gets formatted as:
````haskell
withComments
    :: -- Comment
    Int
    -> -- Comment
    Int
    -> -- Comment
    Int
````

**Comment by brandonchinn178** [comment]
@3kyro thanks! The situation isn't much better on `main`, so it's not too much of a regression
```hs
foo
    :: -- asdf
    Int
    -- bar
    -> Bool
```
But I'll take a look!

3kyro:
> The situation isn't much better on main

Indeed it's not 😞 

Apart from this, the fixes look like a big improvement and I didn't notice any other formatting change regarding leading function arrows.

brandonchinn178:
@3kyro just pushed 3 new commits! should have the comments fixed, and i also found an issue with leading arrows + explicit type signatures for expressions

Issue #233, opened 2022-08-19
Add 'only-args' option to 'function-arrows'


brandonchinn178:
**Describe the option you'd like to add**
#209 added support for leading arrows, but one thing people might prefer is a mixture of leading/trailing, where only the thin arrows are leading. Why?

* `::` is next to function name, for greppability
* Can use trailing haddocks for arguments
* Friendlier diffs when adding explicit `forall` or constraints

**Show a code example exemplifying the different option settings**
```hs
-- trailing
foo ::
  forall a.
  (C1, C2) =>
  Int ->
  String ->
  Bool

-- leading
foo
  :: forall a
  . (C1, C2)
  => Int
  -> String
  -> Bool

-- only-args
foo ::
  forall a.
  (C1, C2) =>
     Int
  -> String
  -> Bool
```

tchoutri:
I would definitely support an option to keep the `::` on the same line as the function name and having aligned leading arrows. Especially for those of us who can't use HLS yet and don't necessarily have a tags program installed.

brandonchinn178:
Is there a reason why you want everything aligned after the function name? The problem with that is the bad diffs when renaming the function. IMO it also doesnt look great to have signatures misaligned across a file

```hs
someLongFuncName :: Int
                 -> String
                 -> Bool

anotherFunc :: Int
            -> String
            -> Bool
```

PR #248, opened 2022-11-01
Add leading-args option for function-arrows

brandonchinn178:
Blocked by #247 
Resolves https://github.com/fourmolu/fourmolu/issues/233


data/fourmolu/function-arrows/output-LeadingArgsArrows.hs
```diff
@@ -0,0 +1,129 @@
...
+
+module Main where
+
+-- | Something else.
+class Bar a where
+    -- | Bar
+    bar ::
+        String
+        -> String
+        -> a
```
brandonchinn178, line 12: The original issue suggested that this first argument would be indented like
```hs
   String
-> String
-> a
```
but it wasn't trivial to implement that exactly, and I don't see it as a big deal. If a first arg becomes the second arg, you have a diff either way:
```diff
-String
+Int
+-> String
 -> String
 -> a

-   String
+   Int
+-> String
 -> String
 -> a
```

brandonchinn178:
cc @tchoutri would love to get your feedback here, given that you commented on the original issue.
cc @georgefst would also be interested in your thoughts

georgefst (approved):
I don't ever intend to use this, so I won't comment on output. But the implementation looks fine.

brandonchinn178:
Merging now, we can improve upon this later. (Specifically, would like to improve #269)

Issue #212, opened 2022-07-08
Converting multiline haddock module description with markup creates parse error


agentultra:
**Is your feature request specific to Fourmolu?**
- [x] Yes, this feature is not relevant to Ormolu

**Describe the current + desired formatting**
```hs
{-|
This example
@
  test
@
-}
```
Gets translated to
```hs
-- |
--This example
--@
--  test
--@
```
Which will result in a parse error on `--@`

I'd recommend it translate to
```hs
-- |
-- This example
-- @
--  test
-- @
```

brandonchinn178:
Thanks for the report! @georgefst, curious what your thoughts are on this. On one hand, it's been bothering me for a while that
```hs
-- | a
-- multiline
-- comment
```
gets converted into
```hs
{- | a
 multiline
 comment
-}
```
with leading spaces, since with single line `-- ` comment style, I don't think of the leading space as part of the comment, but as part of the syntax. But I also get that the GHC parser parses the space as part of the comment content, and we probably don't want to modify what GHC parsed.

georgefst:
> On one hand, it's been bothering me for a while that

Yeah, it annoys me too. I think we can and should fix it.

> But I also get that the GHC parser parses the space as part of the comment content, and we probably don't want to modify what GHC parsed.

I'm pretty sure we already have some sort of `ignoreASTChange` function somewhere (inherited from Ormolu) for special cases where we know we're technically changing the AST, but nobody's likely to care.

georgefst:
As for this issue, I think inserting a leading space on all lines (when there isn't one already) would be totally reasonable. Especially since [#120](https://github.com/tweag/ormolu/pull/809).

PR #255, opened 2022-11-13
Add space after '--' in single-line comments

brandonchinn178:
Fixes https://github.com/fourmolu/fourmolu/issues/212

### DIFF
--- a/src/Ormolu/Utils.hs
+++ b/src/Ormolu/Utils.hs
@@ -76,6 +76,7 @@ splitDocString docStr =
   where
     r =
       fmap (escapeLeadingDollar . escapeCommentBraces)
+        . dropPaddingSpace
         . dropWhileEnd T.null
         . fmap (T.stripEnd . T.pack)
         . lines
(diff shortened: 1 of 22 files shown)


src/Ormolu/Diff/ParseResult.hs
```diff
@@ -125,7 +126,7 @@ matchIgnoringSrcSpans a = genericQuery a
       case cast str1' :: Maybe HsDocString of
         Nothing -> Different []
         Just str1 ->
-          if splitDocString str0 == splitDocString str1
+          if (map T.strip . splitDocString) str0 == (map T.strip . splitDocString) str1
```
brandonchinn178, line 129: We won't care about leading/trailing whitespace in comments in the idempotence test. We should see any unexpected whitespace in test output anyway. If we find idempotence issues later, we could change this to something like
```hs
let normalize s =
      case T.uncons s of
        Just (' ', s') -> s'
        _ -> s
 in if (map normalize . splitDocString) str0 == (map normalize . splitDocString) str1
```

src/Ormolu/Utils.hs
```diff
@@ -85,6 +86,20 @@ splitDocString docStr =
...
+          let leadingSpace txt = case T.uncons txt of
+                Just (' ', _) -> True
+                _ -> False
+              dropSpace txt =
+                if leadingSpace txt
+                  then T.drop 1 txt
+                  else txt
+           in if leadingSpace x
+                then dropSpace <$> xs
+                else xs
```
brandonchinn178, line 102: This is on the Ormolu side. Must've been dropped in a bad merge
https://github.com/tweag/ormolu/blob/54642a7966d1d0787ba84636a2eb52f6729f600e/src/Ormolu/Utils.hs#L89-L102

georgefst: Probably me back when I first implemented multi-line-style haddock. Good find.

georgefst:
Nice! This fixes something that has been minorly annoying me for two years.

georgefst (approved):

PR #263, opened 2022-11-23
Remove `find`-based examples from README

georgefst:


brandonchinn178 (approved):
I would weakly recommend keeping some reference to `git ls-files` around, for people who might not be aware of it, and `fourmolu -i .` would include any stack-work or dist-newstyle directories

Issue #282, opened 2022-12-23
Add "Fourmolu Live"


brandonchinn178:
I really like https://ormolu-live.tweag.io/ because it allows me to try out Ormolu formatting without downloading it. Since it's (presumably) deployed after every push, it also represents the most up-to-date formatting, so users could quickly try it before writing issues, mitigating "this was fixed in the recent version, are you using the latest version" situations. (e.g. #279)

General steps needed:
1. Add `ormolu-live` (renamed to `fourmolu-live`) to `stack.yaml` + `cabal.project`
2. Delete `deploy.sh` and `shell.nix` (and update `DEVELOPER.md` instructions for merging upstream)
3. Make a new `fourmolu.github.io` repo that will host the page with [GitHub pages](https://docs.github.com/en/pages/getting-started-with-github-pages/about-github-pages#types-of-github-pages-sites)
4. Write GitHub actions workflow to build `fourmolu-live` every PR, and, on `main`, push new static files to the `fourmolu.github.io` repo
5. Point repo website to `https://fourmolu.github.io`

Future work:
1. Add configuration section to the page
2. Add docs under `https://fourmolu.github.io/config` that provides pages for each configuration option showing example
    * Generate with `config/Generate.hs`
    * Make sure each option indicates when it was added (new options should say `unreleased`)
    * Add link from main page to config docs
    * Replace configuration section in README with link
    * Update `DEVELOPER.md` + `pr-review-comment.yml` to include updating these docs
3. Update `DEVELOPER.md` to audit these docs every release
    * Replace all `unreleased` tags with new version tag

georgefst:
Yep, this would be pretty cool.

For reference, I've done something similar before [for pretty-simple](https://cdepillabout.github.io/pretty-simple/), implemented in https://github.com/cdepillabout/pretty-simple/pull/116 (styling is still pretty rubbish - I'll get back to that some day...). Notably that has code for drop-downs for selecting enum options from a list, which Ormolu Live does not. In fact, while I'm not necessarily suggesting it's the right approach, it would probably be pretty easy to adapt the pretty-simple code for Fourmolu's purpose. I've thought about abstracting out a library for converting Haskell functions like `SomeSortOfOptionConstraint opts => opts -> Text -> Text` in to web apps.

georgefst:
There is an argument for waiting until GHC's native JS backend has obsoleted GHCJS, so we don't have to faff about with Nix, but this could well take until GHC 9.8.

[...]

PR #283, opened 2022-12-26
Generate PrinterOpts definitions without TH

brandonchinn178:
While investigating #282, I was thinking that Template Haskell might not be great for either GHCJS or the new GHC javascript backend. Plus, Template Haskell is not easily auditable, is difficult to understand, and the code still has a lot of boilerplate (see all the `BijectiveMap` stuff).

Furthermore, if we ever get Fourmolu Live running, I was thinking it'd be cool to move the configuration documentation in the README into the website, and it'd be nice to have one source of truth between the Haskell source code and the website docs.

So I experimented with writing a quick Python script that takes in a YAML file and a Jinja template and generates a Haskell file that's committed to the repo. This is _much_ more auditable, avoids all the negatives of TH, and can generate even more things based off the same YAML file. In this PR, I even had the script generate the in-repo `fourmolu.yaml` file that mimics Ormolu's settings.


brandonchinn178:
@georgefst I think I really like this new approach, but I'd appreciate a second opinion, in case I overlooked anything. Let me know if you have any concerns!

georgefst (approved):
I think this is overall a good idea, and solidly executed, even if it isn't quite the approach I'd take personally (see #287).

brandonchinn178:
@georgefst reimplemented with Haskell. I do like the Python version more, but I don't dislike the Haskell version, and I do agree that it's probably more approachable for contributors.

georgefst (approved):
Nice work. And kudos for the GHC proposals for interpolation and multi-line strings!

PR #287, opened 2023-01-07
Use Dhall instead of Jinja for config code generation

georgefst:
I like the idea of #283, but I have a pretty strong aversion to the whole Python ecosystem.

I actually started off intending to just replace the Python code with Haskell, keeping Yaml and Jinja, but after playing around with it, I just ended up encoding the data, logic and templating all in Dhall. It's a slightly odd language in some ways, but IMO it's fundamentally superior technology to something like Jinja, which feels pretty unprincipled and arbitrary. Besides, as a typed functional language with similar-ish syntax to Haskell, it's likely to feel more familiar to our contributors.

Some comments from the original PR have not been preserved. They mostly documented implicit types which we've now formalised, so I'm not sure they'd serve much purpose. But feel free to re-add anything you think is missing.

I'm actually amazed by the fact that the line count is very similar, given that Dhall is a pretty verbose language (e.g. in requiring types to always be explicit). To me, this indicates that the logic itself is much simpler.


brandonchinn178:
One reason I like Python better is because pretty much all OSes already have Python 3 installed, so a contributor can run `config/generate.sh` and it'll Just Work. And YAML is nice because people are already familiar with YAML. With this, we'd basically require every contributor to install + learn Dhall (considering most PRs are adding a configuration option), which seems like a rather high bar.

In the original PR, a typical contribution would involve editing a YAML file and running a bash script, which is super easy. Considering this workflow is _much_ more common than editing the generation logic, I would rather optimize for that.

One compromise would be to replace Dhall with Haskell (with or without YAML), just building up a string in Haskell without any templating language. My preferences: Python+YAML+Jinja 9/10, Haskell+YAML 5.5/10, Haskell 5/10, Dhall 2/10. I suspect your preferences are the exact opposite of that 😄 

My random personal notes:
* I get that Python + Jinja (+ YAML) can feel icky and too lax, but IMO it's super easy to read, which I think is more important than robustness for this kind of thing
* I personally also like the separation between the templating, the logic, and the data; it's a bit confusing reading through the Dhall to my inexperienced eyes because it's all mixed together

georgefst:
> One reason I like Python better is because pretty much all OSes already have Python 3 installed, so a contributor can run config/generate.sh and it'll Just Work.

That's fair, and probably a pretty big deal in practice. But I'm an idealist and I hate that sort of argument! I tend to prefer to endure some short-term pain in the name of progress.

Also, we _could_ automate the Dhall installation using Cabal. But it seems silly to build from source, seeing as it's already distributed for all the major OSes.

> And YAML is nice because people are already familiar with YAML. With this, we'd basically require every contributor to install + learn Dhall (considering most PRs are adding a configuration option), which seems like a rather high bar.

Well, personally I've never written anything complex in YAML, and the syntax rules for anything non-trivial confuse me every time. And the average contributor only needs to make changes to `data.dhall`, which can likely be understood without knowing much at all about the language.

> In the original PR, a typical contribution would involve editing a YAML file and running a bash script, which is super easy.

Well, it _sounds_ easy, and is once you know what you're doing. But I think the lack of formal structure (types, basically) actually makes it harder to work out what's going on and what's needed. Compare what happens if the contributor tries to use an enum constructor without defining it. In the current implementation, they get a difficult-to-trace Python dict error. Or if they try to use an undefined type, this passes all the way through and only shows up as a type error in the Haskell source. Yes, we could presumably add better error handling in the Python code on an ad-hoc basis, but this is boilerplate-y and silly.

> Considering this workflow is much more common than editing the generation logic, I would rather optimize for that.

I totally agree: realistically, it's very possibly just going to be the two of us editing the non-data parts. But, as stated above, I still think Dhall shines here.

> One compromise would be to replace Dhall with Haskell (with or without YAML), just building up a string in Haskell without any templating language. My preferences: Python+YAML+Jinja 9/10, Haskell+YAML 5.5/10, Haskell 5/10, Dhall 2/10. I suspect your preferences are the exact opposite of that

More or less. Though I think Haskell lacking interpolation and good multi-line strings makes me a bit wary of the middle two.

> I get that Python + Jinja (+ YAML) can feel icky and too lax, but IMO it's super easy to read, which I think is more important than robustness for this kind of thing

It's partly/largely a matter of taste, but I can't agree on the easy-to-read part. Maybe in a trivial purely-aesthetic sense, but not when it comes to ascertaining meaning.

> I personally also like the separation between the templating, the logic, and the data; it's a bit confusing reading through the Dhall to my inexperienced eyes because it's all mixed together

We still have the data separated. The other two could be separated further by abstracting out some functions. I've already done that where I thought it would help, but we could go further.

_Having said all that_, I don't want to force anything you're not happy with, especially as you're by far the more active maintainer these days. And the fact that I'm relying on at least one Dhall feature from a more recent version than the one shipped with the latest Ubuntu kills this for me. It's annoying for CI and a significant chunk of the probable user base. So I'll keep this in the back pocket as something to potentially revisit some day.

brandonchinn178:
Let me prototype something with Haskell first 😄 (between printf and unlines, I think I can get something decent)

georgefst:
Any good reason to delete the branch? I might want to look at this again some day, even just for reference while implementing something else.

brandonchinn178:
The commit is already on github, so youll always be able to access the code (at least online; pretty sure it would let you `git checkout <hash>` locally too)

PR #289, opened 2023-01-21
Implement Fourmolu Live

### LINKED ISSUE #282: Add "Fourmolu Live"

brandonchinn178:
Resolves #282. Adds a top-level `web/` directory that is deployed to `fourmolu.github.io` every merge to `main`.

## Design

Hakyll takes all the files under `web/site/` and compiles them into a statically generated site; at a high level, each Markdown file in `web/site/pages/` is a page in the website (`web/site/pages/foo/bar.md` is served at `/foo/bar`), and all files in `web/site/static/` are copied verbatim.

Aside from the very first page containing the demo, the site is completely workable with just the Hakyll build. Everything is built at build-time, and Bootstrap does the heavy lifting for all dynamic interactions.

### Demo page

The initial page loads `demo.js`, which does two things:
1. registers listeners on all the elements in the demo page
2. starts a web worker that runs as a daemon in the background

Any time an input changes, `demo.js` serializes all the inputs into a JSON object, sends it to the web worker, and displays the output in the relevant HTML elements.

### Web worker

The web worker that `demo.js` starts is implemented in `web/worker/`. We're implementing this as a web worker because it's computationally heavy, so we want to run it in the background, so that the UI remains responsive. The _sole_ purpose of this script is to fetch the WASM binary and convert between the JS and Haskell runtimes. This should be edited very rarely.

We use webpack here because we need a third party dependency to implement WASI (an extension to WASM that allows things like setting environment variables) and we'd like to just serve one JS bundle to run as the web worker. It's annoying to import files in the browser, so webpack allows us to bundle the third party code with the web worker code in one file.

### WASM

The last piece of the puzzle is the WASM binary itself, implemented in `web/fourmolu-wasm/`. This contains a fairly small Haskell script that imports the normal `fourmolu` library and exports the `runFourmolu` function that takes in a JSON string containing the input and config options, and returns a JSON string containing the formatted output, any errors, etc.

This is the most annoying part of all of this, since GHC isn't releasing binaries that can compile to WASM; you have to build from source. I include some instructions in the README, but you shouldn't really need to build this locally. It'll build in CI, and that's the important part.

## Reasonings behind decisions made in this PR

Trying to intentionally use simpler tools, to avoid maintenance burden. Specifically, I wanted to lean on HTML, Javascript, and Markdown, all of which are standard tools that are easy to understand (e.g. as opposed to Reflex or other GHCJS library, which have a bit of a learning curve). Overall, contributing to these docs should be stupidly easy, as we'll be asking all contributors to add docs for their config options.

I also wanted to produce an optimized static site; the only dynamic content on the site is the demo page, with everything else being generated at build time (e.g. formatting the config examples). This minimizes the number of moving parts with a good UX. For the time being, I'm using Hakyll to generate the site, although I'm prototyping a new static site generator that I think would greatly improve the rough implementation here.


brandonchinn178:
FYI @parsonsmatt @georgefst my work so far is deployed on https://fourmolu.github.io! The only thing remaining is a memory issue that comes up after a few formatting calls in the demo, which I'm asking for help on ([GHC issue](https://gitlab.haskell.org/ghc/ghc/-/issues/22947)). But feel free to poke around + provide feedback on the current work!

georgefst:
This looks really cool! I was assuming you'd use GHCJS with Miso or something, but this is a fun proof-of-concept for using the WASM backend.

I'll save reviewing fully until you've implemented [the suggested fix](https://gitlab.haskell.org/ghc/ghc/-/issues/22947#note_481505). And for when I've got the time to set up the toolchain properly. I was hoping the [bindists](https://downloads.haskell.org/ghc/9.6.1-alpha3/) would contain one I could point GHCup to then just use `cd fourmolu && cabal build all -w wasm-ghc-9.6.1-alpha3`. But I guess GHC HQ are focusing on runtime-retargetability rather than ever setting up CI to output WASM-targetting binaries.

brandonchinn178:
Yeah, the key principle I was aiming for was simplicity. Sure, I could have implemented the demo as a full GHCJS app (I couldnt get GHCJS working on my laptop anyway, so this point is mostly moot anyway), but we don't need a full framework for such a simple app. A little HTML + Javascript is much safer (e.g. Miso could release breaking changes) and much lower maintenance (almost zero compilation time, probably an order of magnitude fewer lines of code, conceptually simpler model).

And yeah, GHC isnt releasing WASM binaries. Annoying, but hopefully just a short-term issue. Most of the time, it won't be needed though. Contributors can write docs for their config option and run the static site if they want to preview their docs; they just wont be able to run the demo.

I've implemented the fix, and it does work! My company was acquired, though, so I need to figure out the open source policy with the new company before pushing code changes

georgefst:
> My company was acquired, though, so I need to figure out the open source policy with the new company before pushing code changes

Ah, good luck with that (and if they try to make you sign a contract that restricts or claims ownership of OSS contributions made in your own time, run!).

brandonchinn178:
@georgefst would you be able to take a look at this anytime soon? Would be great to release the next version of Fourmolu alongside a new website

georgefst:
Yes, I've been trying to get to this. Hopefully this Sunday at the latest.

brandonchinn178:
Note to self: use https://github.com/minoki/ghc-docker instead of my hacky docker images

georgefst:
This is really great and I totally agree with the overall direction taken. Unfortunately it's more complex than I expected!

I'm quite happy not to take the time to understand every line. In fact, this is the kind of thing I'd probably allow even as a drive-by contribution, let alone something coming from a maintainer, seeing as it only adds functionality, without affecting the source of the formatter itself. Besides, it doesn't look likely to be too awkward to maintain. However, there are a few things I do intend to review more thoroughly, and I'm sorry to say that I have run out of time for this weekend:
- I would like to make sure that I fully understand the project structure. For example, I'm currently trying to tease HLS in to loading all the source files in the repo correctly, and this is proving somewhat difficult. I think, in general, having multiple `cabal.project`s isn't recommended, although it is possible that it's unavoidable for now due to workarounds needed for the WASM build which we don't want to apply elsewhere. I'm also a bit wary of the symlinks, and might propose that we instead create more internal libraries.
- I want to make sure I can build and run the web app locally. This will hopefully be straightforward, but I've been waiting a long time for all the dependencies to get built!

As for the release, as far as I can tell, nothing else is holding that up, and it would be a shame not to get GHC 9.6 support and other improvements out the door as soon as possible. We've done a good job in the past of following Ormolu releases closely, and I'd like to try to keep that up (I'm aware it's largely my fault for not being very active, and not paying enough attention to realise, until last week, that there hasn't been a release within almost two months of https://github.com/fourmolu/fourmolu/pull/293). So I'd lean towards releasing 0.11 ASAP, and waiting until we're both happy with this PR before separately announcing the web app.

georgefst:
> I want to make sure I can build and run the web app locally. This will hopefully be straightforward, but I've been waiting a long time for all the dependencies to get built!

Ah, it finished just as I posted that! Works nicely, except that I had to run `web/fourmolu-wasm/build.sh` twice after hitting `cp: cannot stat 'HEAD is now at 7745cda WASM-specific fixes'$'\n''HEAD is now at 8cf22af Clean up provider-specific modules'$'\n''/home/gthomas/code/fourmolu/web/fourmolu-wasm/dist-newstyle/build/wasm32-wasi/ghc-9.7.20230313/fourmolu-wasm-0/x/fourmolu-wasm/build/fourmolu-wasm/fourmolu-wasm.wasm': No such file or directory` the first time. Though that's probably an upstream issue anyway. I'm using the Nix flake FWIW.

brandonchinn178:
> So I'd lean towards releasing 0.11 ASAP, and waiting until we're both happy with this PR before separately announcing the web app.

Makes sense to me. I'll release 0.11 now, and we can continue working on this PR.

I have two sources of tension for this PR:
1. I'd really like to get this PR to a state of "we're both enthusiastic about merging this PR". Right now, it seems like you're not quite there yet, and I'd also like to get you 110% on board before merging
2. But also, it'd be nice to get this merged sooner rather than later, so that I don't have to backfill any new options that are merged in the meantime.

That being said, I don't see any PRs coming through lately, so I'm happy to let you take your time reviewing for now, until we see more PRs coming through

brandonchinn178:
Made the release!

>  it would be a shame not to get GHC 9.4 support and other improvements out the door as soon as possible

It seems like 0.10 already had GHC 9.4 support? At least, fourmolu 0.10 was using ghc-lib-parser 9.4

> `cp: cannot stat 'HEAD is now at 7745cda WASM-specific fixes'$'\n''HEAD is now at 8cf22af Clean up provider-specific modules'$'\n''/home/gthomas/code/fourmolu/web/fourmolu-wasm/dist-newstyle/build/wasm32-wasi/ghc-9.7.20230313/fourmolu-wasm-0/x/fourmolu-wasm/build/fourmolu-wasm/fourmolu-wasm.wasm': No such file or directory`

That's a really strange error. It seems to be coming from [this line](https://github.com/fourmolu/fourmolu/blob/7346bdc6d7973552e233007d56ee7aa7bb94038a/web/fourmolu-wasm/build.sh#L21), which means that `wasm32-wasi-cabal list-bin fourmolu-wasm` in this line:

https://github.com/fourmolu/fourmolu/blob/7346bdc6d7973552e233007d56ee7aa7bb94038a/web/fourmolu-wasm/build.sh#L10

is returning the following text to stdout:
```
HEAD is now at 7745cda WASM-specific fixes
HEAD is now at 8cf22af Clean up provider-specific modules
/home/gthomas/code/fourmolu/web/fourmolu-wasm/dist-newstyle/build/wasm32-wasi/ghc-9.7.20230313/fourmolu-wasm-0/x/fourmolu-wasm/build/fourmolu-wasm/fourmolu-wasm.wasm
```

I'm not sure why calling `wasm32-wasi-cabal` would be outputting git information; maybe inspect the `wasm32-wasi-cabal` executable? It should be a script.

georgefst:
> It seems like 0.10 already had GHC 9.4 support? At least, fourmolu 0.10 was using ghc-lib-parser 9.4

My mistake. There _was_ something added since that improved 9.4 support in some way, but I've forgotten the details and it doesn't matter now anyway.

> That's a really strange error. It seems to be coming from this line, which means that wasm32-wasi-cabal list-bin fourmolu-wasm in this line:

Based on the linked issue, it looks like a workaround would be to lower Cabal's verbosity level?

> I think it'd still be nice for the entire web/ project to have a separate cabal.project

You might be right. Another alternative would be a separate repository entirely, but I'm sure that would have its own downsides. I'll need to give this some thought.

> You're just referencing the web/site/src/FourmoluConfig/ symlink, right? Sure, I could make it an internal library. I do like the relative simplicity of pretending the files are just copied over, but I don't have a strong opposition.

In that case, I think we should do it. Sorry for the hassle. But I've just found that a lot of tooling struggles when there's no 1-1 mapping between logical source files and files on disk. Especially when that can lead to files escaping their project's scope.

> @georgefst It seems like the only other blocker is you just understanding the project structure more, right? When do you think you'd be able to continue reviewing? Also happy to jump on a call to walk through the change; feel free to email me.

Essentially, yes. I'll try to make sure I don't hold this up for more than another week. It looks like we'll want to make another release ASAP, given Ormolu's release with GHC 9.6 support etc. So hopefully we'll be able to make the website part of the release this time. I might take you up on that offer of a call, if there's anything complex enough to warrant it. Thanks.

georgefst:
In the meantime, given the upstream merge and new options, could we get this rebased? Thanks.

brandonchinn178:
* Rebased after ormolu merge ✅ 
* Added docs for new options ✅ 
* Changed symlink to internal `fourmolu-config` package ✅ 
* Lowered Cabal verbosity level to workaround `list-bin` issue ✅ 
* Wait for CI to pass ⌛ 
* Wait for final approval ⌛

.github/workflows/pr-review-comment.yml
```diff
@@ -18,6 +18,5 @@ jobs:
 
             Reviewer: Please verify the following things have been done, if applicable.
             - [ ] A file has been added to `changelog.d/`
-            - [ ] "Configuration > Available options" section in `README.md` has been updated
-            - [ ] "Configuration > Specifying configuration" section in `README.md` has been updated
+            - [ ] Docs have been updated in `web/site/config/`
```
georgefst, line 21: This directory does not exist.

web/fourmolu-wasm/fourmolu-wasm.cabal
```diff
@@ -0,0 +1,36 @@
...
+
+executable fourmolu-wasm
+    main-is: Main.hs
+    hs-source-dirs: src
+    other-modules:
+        ForeignUtils
+    default-language: GHC2021
+    ghc-options:
+        -Wall
+        -Werror
```
georgefst, line 14: As with the main package, I would prefer that this were only enabled on CI. Same goes for `fourmolu-web.cabal`.

Although if, for whatever reason, that's difficult, it doesn't matter too much.

web/fourmolu-wasm/src/Main.hs
```diff
@@ -0,0 +1,207 @@
...
+  getString :: Ptr StringWithLen -> IO (Ptr CChar)
+
+foreign export ccall
+  getStringLen :: Ptr StringWithLen -> IO Int
+
+foreign export ccall
+  freeStringWithLen :: Ptr StringWithLen -> IO ()
+
+main :: IO ()
+main = pure ()
```
georgefst, line 51: Is there some quirk of GHC's WASM support that means this needs to be an executable rather than a library?

web/fourmolu-wasm/hie.yaml
```diff
@@ -0,0 +1,4 @@
+cradle:
+  cabal:
+    - path: "."
+      component: "fourmolu-wasm:exe:fourmolu-wasm"
```
georgefst, line 4: These changes are enough to get the code HLS-loadable, which is enough to satisfy me. Though it's not as convenient as it could be, due to HLS generally wanting workspaces to align with `cabal.project`s.

I suspect things _could_ be further improved upstream. See https://github.com/haskell/cabal/issues/5081 for some background.

web/fourmolu-wasm/src/Main.hs
brandonchinn178: Yeah :( WASM forces `main` to exist; I can't get it working without

brandonchinn178:
@georgefst all of that sounds good. After resolving those comments, are we good to go?

fourmolu.cabal
```diff
@@ -127,6 +127,10 @@ library
         scientific >=0.3.2 && <1,
         yaml >=0.11.6.0 && <1
 
+    -- remove after support for 9.8
+    if impl(ghc > 9.6)
+        ghc-options: -Wno-x-partial
```
brandonchinn178, line 132: The WASM build in CI uses GHC HEAD, which has merged the head/tail WARNING change

brandonchinn178:
@georgefst CI finally passed 😤 anything blocking merge?

web/fourmolu-wasm/src/Main.hs
georgefst: Weird. Is there an issue upstream tracking this?

georgefst:
> @georgefst CI finally passed triumph anything blocking merge?

Nope! Thanks for the patience!

georgefst (approved):

web/fourmolu-wasm/src/Main.hs
brandonchinn178: Not sure

config/FourmoluConfig/ConfigData.hs
```diff
@@ -64,6 +65,7 @@ options =
         type_ = "ColumnLimit",
         default_ = HsExpr "NoLimit",
         ormolu = HsExpr "NoLimit",
+        sinceVersion = "0.12.0.0",
```
brandonchinn178, line 68: Ah technically this should've been "Unreleased", and then the release process would've changed this to 0.12. Oh well. Future options should say "Unreleased" in the PR introducing them, per the instructions in `DEVELOPER.md`

PR #292, opened 2023-01-29
Add 'no-align' value to 'in-style' option

### LINKED ISSUE #291: Add `in-style` value to not align let body with patterns
**Describe the option you'd like to add**
This would be an additional value for the `in-style` option; tentative name is `no-align` but this can be confusing with the currently existing values `left-align` and `right-align`, so suggestions welcome.

**Show a code example exemplifying the different option settings**
`in-style: left-align` (current result)
```hs
let Con a b = rhs1
    Con c d = rhs2
in  Con e f
```
`in-style: right-align` (current result)
```hs
let Con a b = rhs1
    Con c d = rhs2
 in Con e f
```
`in-style: no-align` (suggested new value)
```hs
let Con a b = rhs1
    Con c d = rhs2
in Con e f
```

**Motivation**
Short answer: I like my code this way. :)
<details><summary>Long answer: (click to expand, warning: subjective)</summary>
Both existing values align the let body with the patterns in its bindings, which makes no sense to me since the body (an expression) and the patterns are not even in the same syntactic category. Hence visual patterns, which alignment helps the human eye spot, will usually be deceptive instead of clarifying. This is especially true with large left-hand side patterns. Not adding additional spaces around `in` simultaneously solves this problem (by visually setting the body apart) and eliminates some spurious spaces that don't follow indentation levels.
</details>

**Implementation**
Because this would, as far as I can tell, be a very small change, I can implement this in a PR if you wish.

tomsmeding:
This new option value adds no spaces around `in` when unneccessary, resulting in e.g. the following code:

```haskell
let x = 1
    y = 2
in x + y
```

Closes #291. For background and motivation, see there. Note that I am not attached to the name "`no-align`"; better names are welcome.

I slightly reorganised the let-formatting code to centralise the conditions influencing `in`-alignment into a single `case`-expression; I think this makes the logic and the various pieces of data that go into it clearer to follow. But I'm open to any and all suggestions.


brandonchinn178 (approved):
Looks great! Maybe `in-style: minimal`?

tomsmeding:
Thanks @brandonchinn178!
`no-space` feels like a good alternative to me; I leave it to fourmolu maintainers to decide :)
(If you're ambivalent, I think `no-space` is the best option yet)

brandonchinn178:
Great, feel free to update with `no-space` and I can kick off CI!

tomsmeding:
@brandonchinn178 Done! :)

changelog.d/fourmolu-0.10.2.0.md
```diff
@@ -0,0 +1,2 @@
+* Added the `no-space` value for the `in-style` option. See [issue
```
brandonchinn178, line 1: Sorry, small change here. The file should be specific to your change. Can you rename this file to `changelog.d/in-style-no-space.md` or something?

brandonchinn178:
@tomsmeding you need to update the output files for the new config name
```bash
rm data/fourmolu/let-style/output-*-InNoAlign-*.hs
ORMOLU_REGENERATE_EXAMPLES=1 stack test
```

tomsmeding:
@brandonchinn178 Oops, should've tested after I did the rename! Thanks for picking up and merging!

PR #295, opened 2023-02-02
Fix git commit hash in binary

brandonchinn178:
Ormolu changed this in 0.5.3.0, but I forgot to update this in #293


brandonchinn178:
Confirmed that the binary built in CI shows the git commit again

PR #301, opened 2023-03-12
Provide more context on Cabal file parsing errors

zzantares:
When fourmolu fails to parse the Cabal file the user doesn't know what is the actual error as only the cabal filepath is reported. This makes the necessary changes to actually show what were the errors that the cabal file parser encountered.

I made this out of necessity as I didn't know what was the actual problem I was having.

Before:

```
  Parsing this .cabal file failed:
    /path/to/file.cabal
```

After:

```
  Parsing this .cabal file failed:
    /path/to/file.cabal:0:0: Unsupported cabal-version 3.8. See https://github.com/haskell/cabal/issues/4899.
```

_Note_: Removing the `Eq` instance on `OrmoluException` was the easy way to not
  having to implement an orphan `Eq` instance on `PError`. This does not have
  any impact elsewhere as we're actually not making use of the `Eq` instance
  on this exception type.

### DIFF
--- a/src/Ormolu/Exception.hs
+++ b/src/Ormolu/Exception.hs
@@ -16,6 +16,7 @@ import qualified Data.List.NonEmpty as NE
 import Data.Text (Text)
 import qualified Data.Text as T
 import Data.Void (Void)
+import Distribution.Parsec.Error (PError, showPError)
 import GHC.Types.SrcLoc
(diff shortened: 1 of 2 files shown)


brandonchinn178:
This isn't Fourmolu-specific. Mind submitting the PR upstream?

zzantares:
ok opened there https://github.com/tweag/ormolu/pull/999
thank you!

PR #305, opened 2023-03-26
Support automatic line breaking at column limit

CrystalSplitter:
This is a revival of the work done in #102 by @gustavoavena

This feature is implemented as a hack, as described by @georgefst in Issue #71.

This breaks idempotency, which is an unfortunate limitation of this implementation. Open to making fixes for idempotency, but from the comments in #71, this seems to be okay given the utility and the warning.


data/examples/declaration/rewrite-rule/prelude1-line-break-out.hs
brandonchinn178, line 1: Add new tests to `data/fourmolu/` instead, to easily track which tests were added by us and which tests are represented upstream

src/Ormolu/Config/Gen.hs
brandonchinn178, line 1: Did you edit this file manually? As mentioned at the top of this file, this file is autogenerated. Please follow the instructions in `DEVELOPER.md` under the section "Adding a new configuration option"

tests/Ormolu/PrinterSpec.hs
brandonchinn178, line 1: I don't think this file needs to/should change. All the files in `data/` should be formatted with Ormolu's format or Fourmolu's default format.

README.md
```diff
@@ -87,6 +87,8 @@ respectful: true
 fixities: []
 unicode: never
 single-constraint-parens: always
+# Max number of characters in a line. Disclaimer: enabling this breaks idempotence in a few cases.
+column-limit: no-limit
```
brandonchinn178, line 91: Nit: can we make the `NoLimit` option `column-limit: none`? Reads a little better / less redundant to me.

brandonchinn178 (changes requested):

data/examples/declaration/rewrite-rule/prelude1-line-break-out.hs
CrystalSplitter: Can do!

src/Ormolu/Config/Gen.hs
CrystalSplitter: Almost all the code here is rebased and tweaked code from #102 with line-by-line fixes to get everything to work, rather than a complete reimplementation of the work.

I'll scrap the changes to this file and redo them.

tests/Ormolu/PrinterSpec.hs
CrystalSplitter: Is there a way to specify tests for alternative configs? We're not modifying the default behaviour, but it would still be nice to have tests for this. Otherwise there's no reason to add new tests--no behaviour would be changing.

README.md
CrystalSplitter: Sure, pretty easy fix!

tests/Ormolu/PrinterSpec.hs
brandonchinn178: Yes, following the instructions in `DEVELOPER.md` goes through all of that

src/Ormolu/Config/Gen.hs
CrystalSplitter: Done!

data/examples/declaration/rewrite-rule/prelude1-line-break-out.hs
CrystalSplitter: Done!

tests/Ormolu/PrinterSpec.hs
CrystalSplitter: Done, though I still had to make some modifications to how we run `TestGroup`s.

README.md
CrystalSplitter: Done!

data/fourmolu/column-limit/input.hs
```diff
@@ -0,0 +1,82 @@
...
+    let aVeryLongLine = ["list one", "list one", "list one", "list one", "list one", "list one"]
+    in undefined
+
+longFunction3 :: String -> String -> String -> Maybe Int -> Maybe Int -> Maybe Int -> String
+longFunction3 veryLongArg1 veryLongArg2 veryLongArg3 veryLongArg4 veryLongArgument5 veryLongArg6 = undefined
+
+
+-- For reference, this line had exactly 80 characters -------------------------|
+
+-- ----------------  Know limitation: idempotence is broken ----------------
```
brandonchinn178, line 55:
```suggestion
-- ----------------  Known limitation: idempotence is broken ----------------
```

src/Ormolu/Config/Gen.hs
```diff
@@ -442,3 +458,22 @@ instance PrinterOptsFieldType SingleConstraintParens where
...
+      either Aeson.parseFail pure $
+        parsePrinterOptType (Text.unpack s)
+
+instance PrinterOptsFieldType ColumnLimit where
+  parsePrinterOptType =
+    \s ->
+      case s of
+       "none" -> Right NoLimit
+       other ->
+          case (readEither other :: Either String Int) of
```
brandonchinn178, line 474: Nit: does this require the explicit type annotation? Can we remove it?

tests/Ormolu/Config/PrinterOptsSpec.hs
```diff
@@ -162,28 +170,40 @@ singleTests =
...
-          testCaseSuffix = suffix1
+          testCaseSuffix = suffix1,
+          checkIdempotence = True
+        },
+      TestGroup
+        { label = "column-limit",
+          testCases = [NoLimit, ColumnLimit 80, ColumnLimit 100],
+          updateConfig = \columnLimit opts -> opts {poColumnLimit = pure columnLimit},
+          showTestCase = show,
+          testCaseSuffix = suffix1,
```
brandonchinn178, line 205: Nit: Would be great to avoid spaces in the filename. Thoughts on `output-limit=none.hs` + `output-limit=100.hs`?

brandonchinn178 (approved):
Small nits, let me know if you agree, or I can merge as-is

src/Ormolu/Config/Gen.hs
CrystalSplitter: This was leftover from some type debugging. Thanks for the catch. We can remove it.

tests/Ormolu/Config/PrinterOptsSpec.hs
CrystalSplitter: Yeah, I don't like space either. Personally somewhat hesitant on `=`, since that'll also need escaping in many shells.

Any opinions on:

* `output-limit-none.hs`/`output-limit-100.hs`?
* `output-limit-eq-none.hs`/`output-limit-eq-100.hs`?

I favour the first, but anything that doesn't need escaping is good to me.

CrystalSplitter: Ah, I see that the `=` is already common practice in `respectful-module-where`. I can do that.

brandonchinn178: oh feel free to not use `=`. Just out of curiostiy, what escaping issues are there? I don't think there'd be an issue with the variable assignment `=`.

Of the two options you listed, I like the non-`eq` one

CrystalSplitter: > I don't think there'd be an issue with the variable assignment =.

I can only think really think of a variable assignment issue if you're trying to execute a command which contains `=` in the name. e.g.

```bash
# Bash/ZSH
$ echo 'echo RAN' > ~/bin/hello=world && chmod u+x ~/bin/hello=world
$ hello=world  # Variable assignment silence
$ 'hello=world'  # Actually executing it
RAN
```

That's not the case here, but it's always something I'm a little wary of. The other thing that comes to mind is the autocomplete will insert `\` prior to the filenames, but that's more an annoyance than anything actually an issue.

There are also commands come to mind which use `=` for opt assignment in shell commands (particularly ones that don't lead with `-`). This can cause confusion as to whether it's a opt or a filename. Some Golang programs I've worked with do that, but I think that's a _really_ niche case.

---

I've already implemented the `=` filename version, which I think is probably fine?

src/Ormolu/Config/Gen.hs
CrystalSplitter: Done.

data/fourmolu/column-limit/input.hs
CrystalSplitter: Had to regenerate. Done.

tests/Ormolu/Config/PrinterOptsSpec.hs
CrystalSplitter: Done with `=`.

src/Ormolu/Config/Gen.hs
```diff
@@ -442,3 +458,22 @@ instance PrinterOptsFieldType SingleConstraintParens where
...
+    \s ->
+      case s of
+       "none" -> Right NoLimit
+       other ->
+          case readEither other of
+            Left _ -> Left . unlines $
+              [ "unknown value: " <> show s,
+                "Valid values are: \"no-limit\", or an integer"
+              ]
+            (Right otherInt) -> Right . ColumnLimit $ otherInt
```
brandonchinn178, line 479: Nit: didn't catch this earlier, but this could also be

```hs
case s of
  "none" -> ...
  _ | Just x <- readMaybe s -> ...
  _ -> ...
```

brandonchinn178 (approved):
One last nit. Feel free to ignore.

src/Ormolu/Config/Gen.hs
CrystalSplitter: Done. Had to modify the import list (`readMaybe` wasn't available).

CrystalSplitter:
Hold on, parsing issue.

> Error in $['column-limit']: parsing ColumnLimit failed, expected String, but encountered Number.

Maybe have broken it up somewhere in my fixups...

brandonchinn178:
@CrystalSplitter the adtParseJson implementation, I think.

Also, the tests apparently dont test that function... I'll have to think about how best to test it

CrystalSplitter:
The underlying issue here is that column limit can both be a `String` (`none`) and an `Int` during parsing.

Some solutions:

* A real quick solution to this would be that 0 or lower is `none`. This is originally what it did at the very beginning, but it was requested that Gustavo change that. I think this option is fine, though weird from a user perspective.
* Another solution is make a somewhat hacky parser that returns both the string `"none"` and an integer as a raw string. This is fine from a user perspective but weird for implementation.
* And another solution is to write in the documentation that the integer has to be wrapped in quotes. I dislike this option.

brandonchinn178:
Or just parse it directly?

```hs
parseJson = \case
  String "none" -> pure NoLimit
  Number x | Right i <- floatingOrIntegral x -> pure $ ColumnLimit i
  _ -> fail "Invalid ColumnLimit"
```

It'll slightly duplicate the adtParseText implementation, which I think is fine.

CrystalSplitter:
Done, but I had to include `scientific` in the depends. I noticed that there was a test that broke with that, so I patched that as well. I'm unsure if that was the right move, though.

Issue #307, opened 2023-03-30
Add a cli flag to print currently effective options


jhrcek:
Following up on discussion in https://github.com/fourmolu/fourmolu/issues/306#issuecomment-1489651160
It would be nice if there was a command line flag that you could use to
1. print default configuration file to std out (something like [stylish-haskell](https://github.com/haskell/stylish-haskell#configuration) which allows you to dump default config options to file `stylish-haskell --defaults > .stylish-haskell.yaml`)
This file would contain the default settings + comments showing potential alternatives for each option. I would find it more ergonomic than what I currently do, namely going to github readme and copy-pasting example config into my projects.

2. another potentially useful thing could be to print "currently resolved settings" where you'd see how various defaults get combined with configs from your home (~/fourmolu.yaml) and or from your project (.../your-project/fourmolu.yaml).

Just suggesting potential names for 1 and 2. Feel free to suggest better flag names :smile: 
```bash
fourmolu --default-options
fourmolu --effective-options
```

brandonchinn178:
Makes sense to me! We should generate the default YAML file in `Ormolu.Config.Gen` (using similar logic to the CLI description text for the comments for each option) and just `putStrLn` if we get that flag in `Main.hs`.

Bike-shedding: I think an explicit `fourmolu --print-defaults` would be nice, and maybe `fourmolu --print-config` for the current settings?

PR welcome!

brandonchinn178:
oh interesting. Yeah `--print-defaults` might be simpler, since we already generate `fourmolu.yaml` (containing the ormolu configuration). So it's just a matter of generating another one with the default options and showing it.

`--print-config` is a bit harder because we'd have to add a `ToJSON` instance for `PrinterOptsTotal` (and then we could just do `Yaml.encode`). RE: format, I'd say `--print-config` should just output a YAML file, optionally with comments about provenance (although we don't store that information right now, so let's leave that out for now)

georgefst:
It might be worth updating the title or OP here to reflect what remains to be done after #308 (i.e. `--print-config`).

brandonchinn178:
#337 adds the configuration to the `--debug` flag. Maybe that's good enough for this?

brandonchinn178:
@jhrcek `--debug` now includes the configuration being used. Is that sufficient?

jhrcek:
@brandonchinn178 Thank you, sounds good enough for my purposes :smile:  Feel free to close this issue.

PR #308, opened 2023-04-01
Add --print-defaults flag to print default fourmolu.yaml with comments

jhrcek:
First attempt at implementing https://github.com/fourmolu/fourmolu/issues/307

At first I thought I could add ToJSON and just print the `defaultPrinterOpts` serialized to Yaml.
This has the downside of not telling you what are other valid values for each option. And I didn't find a way to *render* comments within Yaml using yaml library.

To get access at the info on valid values + also help text for each option I tried another approach:
I extended `Generate.hs` to assemble output String corresponding to valid yaml from pieces already present in `ConfigData.hs`. I added the resulting String to Gen.hs.
But it feels a bit hacky. Let me know what you think or if you have any ideas how to do it in a less hacky way.

This is the result it produces:
```bash
$ stack --stack-yaml stack-ghc-9.2.yaml exec fourmolu -- --print-defaults
# Number of spaces per indentation step
indentation: 4

# Styling of arrows in type signatures (choices: trailing, leading, or leading-args)
function-arrows: trailing

# How to place commas in multi-line lists, records, etc. (choices: leading or trailing)
comma-style: leading

# Styling of import/export lists (choices: leading, trailing, or diff-friendly)
import-export-style: diff-friendly

# Whether to full-indent or half-indent 'where' bindings past the preceding body
indent-wheres: false

# Whether to leave a space before an opening record brace
record-brace-space: false

# Number of spaces between top-level declarations
newlines-between-decls: 1

# How to print Haddock comments (choices: single-line, multi-line, or multi-line-compact)
haddock-style: multi-line

# How to print module docstring
haddock-style-module: null

# Styling of let blocks (choices: auto, inline, newline, or mixed)
let-style: auto

# How to align the 'in' keyword with respect to the 'let' keyword (choices: left-align, right-align, or no-space)
in-style: right-align

# Output Unicode syntax (choices: detect, always, or never)
unicode: never

# Give the programmer more choice on where to insert blank lines
respectful: true

# Fixity information for operators
fixities: []

# Whether to put parentheses around a single constraint (choices: auto, always, or never)
single-constraint-parens: always

# Max line length for automatic line breaking
column-limit: none

```


tests/Ormolu/Config/OptionsSpec.hs
```diff
@@ -44,6 +44,25 @@ spec =
 
         withCLI <- readProcess fourmoluExe ["--indentation=4", hsFile]
         withCLI `shouldBe` indented4
+
+    it "prints defaults to stdout" $ \fourmoluExe -> do
+      readProcess fourmoluExe ["--print-defaults"]
+        `shouldReturn` "indentation: 4 # Number of spaces per indentation step\n\
```
brandonchinn178, line 50: hm I'm not fond of the fact that this test would have to be updated for every new option. Can you just check that the exit code is success and maybe just the first couple lines in the output?

config/Generate.hs
```diff
@@ -234,6 +239,27 @@ fourmoluYaml = unlines $ header <> config
         | Option {..} <- options
       ]
 
+-- | Default fourmolu config that can be printed via `fourmolu --print-defaults`
+fourmoluYamlFourmoluStyle :: String
+fourmoluYamlFourmoluStyle = unlines config
+  where
+    config =
+      [ printf "%s: %s # %s" name (hs2yaml type_ default_) (getComment opt)
```
brandonchinn178, line 247: Can we show the options like

```yaml
# description
indentation: 4

# next description
newlines-after-decls: 1
```

tests/Ormolu/Config/OptionsSpec.hs
jhrcek: Checking that exit code is success is already done in readProcess test utility:
https://github.com/fourmolu/fourmolu/blob/fba2c49b70bbcfddd99a89ba91060dc400aa2b7a/tests/utils/IntegrationUtils.hs#L22

But I agree with only checking first few lines only

config/Generate.hs
jhrcek: Ok, done.

src/Ormolu/Config/Gen.hs
```diff
@@ -488,3 +489,6 @@ instance PrinterOptsFieldType ColumnLimit where
             [ "unknown value: " <> show s,
               "Valid values are: \"none\", or an integer"
             ]
+
+defaultPrinterOptsYaml :: String
```
jhrcek, line 493: I have some second thoughts about rendering the defaults as a huge one-line string into this file.

What would you say to this different approach:
let Generate.hs render the default options into a file, like data/fourmolu/fourmolu.yaml.defaults

and then just render the following code here:
```haskell
defaultPrinterOptsYaml :: String
defaultPrinterOptsYaml = unpack $(embedFile "data/fourmolu/fourmolu.yaml.defaults")
```

Maybe it would improve reviewability (you'd see actual line changes in the generated defaults yaml file).

brandonchinn178: The whole point of this file is to avoid Template Haskell :)

I'm not terribly concerned about this, but you could generate an `unlines` list if you want?

If only we had multiline strings 🙃 (https://github.com/ghc-proposals/ghc-proposals/pull/569)

jhrcek: I'm not a fan of TH myself :)
let me think about how to make this in the least painful way.

Another thin I don't like about the result is the "null" for haddock-style-module, as well as missing list of valid values.
Let's see if I can improve that somehow
Screenshot from 2023-04-02 09-19-06

brandonchinn178: For the record, avoiding TH is not just preference, it's also to enable running Fourmolu in the browser, since the WASM backend for ghc doesnt support TH yet.

I don't know if we need to spend too much effort improving the descriptions, since we'll hopefully soon have a nice docs site (https://github.com/fourmolu/fourmolu/pull/289). When I get that merged, we could just link to it from the default YAML file.

jhrcek: Ok, I'm ignoring the descriptions, and I went with your suggestion of using unlines on list of lines.
Makes the Generator.hs a bit more complicated, but the improved "reviewability" of changes in the Gen.hs module is worth it IMO.

config/Generate.hs
```diff
@@ -234,6 +244,27 @@ fourmoluYaml = unlines $ header <> config
...
+
+    renderEnumOptions enumOptions =
+      renderList [opt | (_, opt) <- enumOptions]
+
+    getComment Option {..} =
+      let help = fromMaybe description (cliHelp cliOverrides)
+          choicesText =
+            case type_ `Map.lookup` fieldTypesMap of
+              Just FieldTypeEnum {enumOptions} ->
+                printf " (choices: %s)" (renderEnumOptions enumOptions)
```
brandonchinn178, line 264: Nit: could this just be
```suggestion
                printf " (choices: %s)" (renderList $ map snd enumOptions)
```
?

tests/Ormolu/Config/OptionsSpec.hs
```diff
@@ -44,6 +45,11 @@ spec =
 
         withCLI <- readProcess fourmoluExe ["--indentation=4", hsFile]
         withCLI `shouldBe` indented4
+
+    it "prints defaults to stdout" $ \fourmoluExe -> do
+      stdOutput <- readProcess fourmoluExe ["--print-defaults"]
+      -- Only check prefix of the output, so we don't have to update the test with every new option added
+      stdOutput `shouldSatisfy` isPrefixOf "# Number of spaces per indentation step\nindentation: 4\n"
```
brandonchinn178, line 52: Mentioning this for posterity: we could do `stdOutput `shouldBe` fourmoluYamlFourmoluStyle`, but I actually like it this way, with a literal string showing a little bit what it should look like. It makes the test more robust, e.g. if `fourmoluYamlFourmoluStyle` was accidentally generated as an empty string. It also makes the test more meaningful, as opposed to just ensuring we hooked up everything correctly

brandonchinn178 (approved):
Looks great, thanks! Some nits you are free to include, or I can merge without

config/Generate.hs
jhrcek: Applied.

config/Generate.hs
```diff
@@ -203,6 +209,9 @@ configGenHs =
     renderEnumOptions enumOptions =
       renderList [printf "\\\"%s\\\"" opt | (_, opt) <- enumOptions]
 
+    renderMultiLineStringList multiLineStr =
+      unlines $ zipWith (\c str -> c : ' ' : show str) ('[' : repeat ',') (lines multiLineStr)
```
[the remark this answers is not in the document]
jhrcek: Applied.

tests/Ormolu/Config/OptionsSpec.hs
jhrcek: Yes, I considered it, but went with hardcoded string literal exactly for this reason, to have check independent of the String being generated.

README.md
```diff
@@ -92,6 +92,12 @@ single-constraint-parens: always
 column-limit: none
 ```
 
+Starting with fourmolu 0.12.0.0 you can generate the default configuration file like this:
+
+```console
+$ fourmolu --default-config > fourmolu.yaml
```
ncfavier, line 98: I think you meant `--print-defaults` here

brandonchinn178: Yes, thank you! Feel free to put up a PR, or I can update it when I'm free

PR #310, opened 2023-04-03
Merge ormolu 0.6

brandonchinn178:
Big thing in this update is GHC 9.6 support.

One notable update is Ormolu stopped generating any caret haddocks (https://github.com/tweag/ormolu/pull/993), whereas we still want caret haddocks when generating function types with leading arrows. So what I did is (temporarily) remove the pipe/caret distinction in `p_hsType`, then readded it.

Will release 0.12 after this merges

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,9 +14,9 @@ jobs:
     strategy:
       matrix:
         stack_yaml:
-          - stack-ghc-9.0.yaml
           - stack-ghc-9.2.yaml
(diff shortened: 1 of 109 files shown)


georgefst:
Thanks as ever for this unglamorous work! I guess we may as well add the just-released 0.6.0.1 in this PR as well. It contains the important-looking fix https://github.com/tweag/ormolu/pull/1009.

brandonchinn178:
😉 yup will do that after work

PR #318, opened 2023-04-17
Fix fourmolu demo for new ormolu

brandonchinn178:
The new ormolu merge [changed](https://github.com/tweag/ormolu/pull/991) how the fixity database is loaded (preinitialized with wizer instead of fetched at runtime). If you use any infix operators at fourmolu.github.io, it'll show an error.

@georgefst I can't compile this right now because `rts_clearMemory` was added [2 months ago](https://gitlab.haskell.org/ghc/ghc/-/commit/9ca51f9e84abc41ba590203d8bc8df8d6af86db2) while my build of GHC wasm is too old, and I'm getting a weird error about `tzset` being undeclared when rebuilding GHC. I'll try to get it working on my end, but if you can get it working, feel free to test/fix/merge.

### DIFF
--- a/web/README.md
+++ b/web/README.md
@@ -13,27 +13,29 @@ This directory contains files that are served at `fourmolu.github.io`.
 
 ### Build locally
 
-To build locally, follow the steps at https://gitlab.haskell.org/ghc/ghc-wasm-meta to install GHC with the WASM backend. Then run the following steps:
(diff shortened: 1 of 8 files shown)


brandonchinn178:
Ok just downloaded the `fourmolu-wasm.wasm` artifact from CI and tested that it works locally

PR #319, opened 2023-04-17
Fix column-limit input

### LINKED ISSUE #317: Fix column-limit input in demo
The Fourmol demo currently shows the `column-limit` option as a dropdown with "none" or "Any non-negative integer". It's done this way so that the `column-limit` page looks like

Screen Shot 2023-04-16 at 3 30 56 PM

which is what we want it to look like. So we should change the dropdown in the demo while keeping this table the same

brandonchinn178:
Resolves #317 

@georgefst Can't test right now because of #318. Should hopefully work as-is, but don't know for sure. If you can get that working, would be great if you can test this change with the `data/fourmolu/column-limit/input.hs` text + changing the `column-limit` input in the demo

### DIFF
--- a/config/FourmoluConfig/ConfigData.hs
+++ b/config/FourmoluConfig/ConfigData.hs
@@ -45,6 +45,7 @@ data HaskellValue
   | HsInt Int
   | HsBool Bool
   | HsList [HaskellValue]
+  deriving (Eq)
 
 allOptions :: [Option]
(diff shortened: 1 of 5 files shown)


brandonchinn178:
Validated that this works locally, using the binary in #318. I'm gonna merge this now, to expedite the release, but we can make changes after the release.

PR #320, opened 2023-04-20
Demo improvements

brandonchinn178:


web/site/static/demo.js
```diff
@@ -129,6 +135,28 @@ function main() {
...
+        MonadLogger m)
+        => Int -> m ()
+    loadPerson id = runQuery queryText [SqlInt id]
+      >>= toPerson
+      where queryText = "SELECT * FROM person WHERE id = ?"
+            toPerson rows =
+                case rows of
+                  [ [ SqlString name, SqlInt age ] ] -> pure Person { .. }
+                  _->logAndFail $ "Invalid result: "++show rows
+  `.replace(/^[ ]{4}/gm, '').trim()
```
georgefst, line 159: I'm guessing this example is intended to show off as many Fourmolu features as possible? I had thought that it would be nice to have one which shows some difference for every config option. Or a set of examples like https://cdepillabout.github.io/pretty-simple, which took them from existing test cases.

georgefst (approved):

web/site/static/demo.js
brandonchinn178: It's meant to show a reasonable sample of Fourmolu examples. I don't think we should make the example representative of every single configuration option; that's what the configuration pages are for.

georgefst: Fair enough. I think it is nice to have something more interactive. But I'm really not bothered about it. And it's possible that it would have to be a very big code sample to cover everything.

brandonchinn178: I didn't spend too much time coming up with this example; I just wanted to get the ball rolling. If you wanna spend a couple minutes to come up with a better example, I would happily replace it. My only preference would be to keep it under ~100 lines.

georgefst: Okay, I think this covers everything:

```hs
{-# LANGUAGE UnicodeSyntax #-}
-- | This is the
-- person module.
module Person (Person(Person,
name),
loadPerson)where
import Control.Monad
    (void,unless,when)
data Person = Person
  {name::String,
   age::Int}
loadPerson ::
    (MonadIO m)
    => Int -> m ()
loadPerson id = runQuery queryText [SqlInt id]
  >>= toPerson
  where queryText = "SELECT * FROM person WHERE id = ?"
        toPerson rows =
            case rows of
              [ [ SqlString name, SqlInt age ] ] -> pure Person { .. }
              _-> let s = "Invalid result: "++show rows in logAndFail s
```

It's slightly less useful than I imagined, since we lose the scrollbar position in the output when we change an option, meaning we often miss what's just changed. But I still think it's probably worth having.

georgefst:
I've just noticed that the web app's initial values for some options are not the same as Fourmolu's defaults:
- `import-export-style`
- `haddock-style`
- `in-style`
- `single-constraint-parens`
- `unicode`

Is there any reason for this?

brandonchinn178:
@georgefst fixed the defaults issue, the scrollbar issue, and updated the example 😄 Will merge shortly

PR #322, opened 2023-05-10
Add option to preserve the 'monkey-head operator' (:[])

expipiplus1:


brandonchinn178:
@expipiplus1 It looks like you changed the generated files manually. Can you make sure to update the files in `config/`? It also looks like you need to reformat your changes.

More functionally, my concern with this is that it seems a bit bespoke, putting logic for just one hardcoded special case. I wonder if there's a way to generalize this in the configuration (e.g. `operator-sections: ['(:[])', '(++[])']`), or if not, if this specific use-case is broad enough to warrant a special case. Thoughts @georgefst?

expipiplus1:
Yeah, I fully admit that this is trivial and niche! It was just a rabbit-hole itch when I formatted some old source and the monkey friends disappeared :) 

If there's a nice way to parse this list of operator sections to keep "dense" then I'd be happy to change this to use that

brandonchinn178:
Haven't thought it through, but maybe some idea of using ghc-lib-parser to parse each string to get out the operators used and then compare it with the operators in the formatter

expipiplus1:
Some stats from Hackage, monkey head is several times more popular than the alternative:

```
$ ag --literal '(:[])' --stats-only
3130 matches
1855 files contained matches

$ ag --literal '(: [])' --stats-only
523 matches
363 files contained matches

$ ag --literal ':[]' --stats-only
32709 matches
3467 files contained matches

$ ag --literal ': []' --stats-only
10133 matches
2313 files contained matches
```

expipiplus1:
> @expipiplus1 It looks like you changed the generated files manually. Can you make sure to update the files in `config/`? It also looks like you need to reformat your changes.

Thanks, done

brandonchinn178:
@expipiplus1 This looks great, but I'm still wary of such a specific option. What do you think about a more general option, like:
```yaml
compact-sections:
  - (:[])
  - (++[])
```
As I mentioned in a previous comment, you might be able to leverage ghc-lib-parser to parse the expression

georgefst:
A generalised version of this would be cool. I've recently wanted this for a little tennis scoring DSL: `4 – 6, 7 – 5, 6 – 3, 6 – 4` just looks wrong, compared to `4–6, 7–5, 6–3, 6–4`.

PR #323, opened 2023-05-14
Relax upper bound to accommodate ansi-terminal-1.0

mpilgrem:
`ansi-terminal-1.0` is released on Hackage.

See also https://github.com/commercialhaskell/stackage/issues/6976.

Tested by building with Stack >= 2.9.3 and `stack-ghc-9.2.yaml` with additions:
~~~yaml
extra-deps:
  - ansi-terminal-1.0
  - ansi-terminal-types-0.11.5

allow-newer: true
allow-newer-deps:
- ansi-wl-pprint
~~~

### DIFF
--- a/fourmolu.cabal
+++ b/fourmolu.cabal
@@ -107,7 +107,7 @@ library
         Cabal-syntax >=3.10 && <3.11,
         Diff >=0.4 && <1.0,
         MemoTrie >=0.6 && <0.7,
-        ansi-terminal >=0.10 && <1.0,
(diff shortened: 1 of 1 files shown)


brandonchinn178:
Thanks @mpilgrem. Note that there's not much rush on this, since fourmolu 0.12 requires ghc-lib-parser 9.6, so fourmolu wont work with Stackage even if this merges. Until Stackage upgrades ghc-lib-parser, I'm inclined to just wait for ormolu to bump this instead of cherry picking this fix to fourmolu

PR #364, opened 2023-09-03
Add ability to specify presets

### LINKED ISSUE #300: Add preset options
Add a new flag/config option `--preset` that specifies the default values of unspecified options. This enables certain, blessed option sets for various use cases. Possible presets:

* default: The current default Fourmolu options (also the default `preset` if none is specified)
* ormolu: Options that emulate Ormolu settings
* diff-friendly: Options that aim to be as diff-friendly as possible
* aligned: Options that aim for everything to be as aligned as possible

If an option isn't directly related to the preset (e.g. indentation for diff-friendly), it'll be set the same as `default`.

Optionally, we could also allow `preset` to be a list, with latter preset values overriding earlier presets.

"Fourmolu already uses diff-friendly formatting by default, what's the point of the diff-friendly preset?" As we've been adding more options, we (the maintainers) have been pushing for option defaults to not change current formatting, to prevent breaking current formatting and maintain some notion of stability. If we add a new `foo` option tomorrow with a diff-friendly setting, we'd want the default to not change existing formatting (or at least have it be opt-in, a la `let-style: auto`). But a user might explicitly want to opt in to the maximally diff-friendly setting, even if formatting changes over an upgrade. So with presets, we'd be able to say that `default` has no (or very rarely) formatting changes on upgrade (aside from changes we get from ormolu), and other presets would get the cutting edge updates.

Other notes:
* Adding a new configuration would now need to specify the default for fourmolu (as non-breaking as possible), the setting to emulate ormolu, and also the setting for all other presets.
* We could also add a `edge` preset that contains new `default` values for the next major Fourmolu version bump (if we ever wanted to do so)

Related: #128

brandonchinn178:
Resolves #300 

Adds a `preset` option with built-in options `fourmolu` and `ormolu`, plus the ability to specify a URL to download a fourmolu config file.

Probably easiest to review commit-by-commit. There's a lot of refactoring, to separate options used for PrinterOpts and other options that are not in PrinterOpts (`fixities`, `reexports`, `preset`).

Option settings for Fourmolu/Ormolu presets are now broken out into a separate table:
Screenshot 2023-09-03 at 12 09 55 AM


brandonchinn178:
@georgefst any thoughts on this at all?

PR #379, opened 2023-10-25
Fix building WASM in CI

brandonchinn178:


brandonchinn178:
Doing this in #378, since it's easier to do with the switch to ghc-lib-parser 9.8

PR #391, opened 2023-12-04
Fix the order in which the configurations are applied

### LINKED ISSUE #390: Configurations are applied in the wrong order
**Is your bug specific to Fourmolu?**
- [x] Yes, I've verified that Ormolu does not have this bug

**Did you try it on the web app?**
- [ ] Yes, I've verified that https://fourmolu.github.io can reproduce this bug

**Did you search for existing issues?**
- [x] Yes, there are no open or closed issues related to my issue

**Describe the bug**

The Haddock documentation of `resolvePrinterOpts` states the following:
> Apply the given configuration in order (later options override earlier).

However, this is not actually correct as it is currently implemented:

```
ghci> first :: PrinterOptsPartial = mempty {poIndentation = Just 2}
ghci> second :: PrinterOptsPartial = mempty {poIndentation = Just 4}
ghci> poIndentation $ resolvePrinterOpts [first, second]
Identity 2
```

ozkutuk:
Fixes #390

The documentation of `resolvePrinterOpts` states that "later options override earlier", but it is not the case. This commit fixes the application order such that it is inline with the documentation.

### DIFF
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -100,8 +100,8 @@ mkConfig path Opts {optQuiet, optConfig = cliConfig, optPrinterOpts = cliPrinter
     cliConfig
       { cfgPrinterOpts =
           resolvePrinterOpts
-            [ cliPrinterOpts,
-              cfgFilePrinterOpts fourmoluConfig
+            [ cfgFilePrinterOpts fourmoluConfig,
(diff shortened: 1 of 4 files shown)


brandonchinn178 (approved):
Thank you!

PR #397, opened 2024-02-16
Merge ormolu 0.7.4.0

brandonchinn178:


brandonchinn178:
cc @kostmo @Kleidukos
Fourmolu will soon be released with breaking changes to formatting. These changes are coming in from Ormolu, but overall make the styling much better IMO. Most of the changes fix if/then/else blocks being an indentation behind

Kleidukos:
@brandonchinn178 Thanks for the warning. I find this forced change a bit problematic:

```diff
-data SectionDescr a = forall b.
+data SectionDescr a
+  = forall b.
```

We'd be grateful for a flag to control this behaviour, because that's going to impact **a lot** of code. :)

Kleidukos:
Of course please do ask for help if needed, I am not considering the idea that you'd do it alone.

brandonchinn178:
@Kleidukos can you elaborate by "a lot of code"? I only see two instances of it changing in the cabal diff.

As I've mentioned in the past, I'm hesitant about adding flags for every major change Ormolu makes because each flag = complexity and increases possibility of merge conflicts. It's also a huge improvement if you have constructors with long names or pragmas.

Issue #396, opened 2024-02-05
specify alternate configuration file


teto:
I would like to embed the configuration file with ormolu so that the same config can be used in various projects.

we provide fourmolu via nix and one of the common ways to do this in nix is to wrap the executable so:
1/ Ideally I would create a fourmolu-new wrapper that calls `fourmolu --config /nix/store/....-config`
2/ One solution is to create a wrapper fourmolu-new that  calls fourmolu with options passed on the command line. But I prefer to edit the file with comments rather than the command line
3/ My other idea since fourmolu looks for a config in XDG_CONFIG_HOME is to change the value of this variable to a folder in the nix store with the config. It's convoluted.

Is such a command line option desirable ? what do you suggest ?

brandonchinn178:
Sure, we could add a `--config` option! Feel free to open a PR!

teto:
thank you \o/

PR #399, opened 2024-02-21
support passing a config file as a CLI argument

teto:
- **add flake**
- **Load config-file from an alternate path**
Attempt at addressing https://github.com/fourmolu/fourmolu/issues/396

### DIFF
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -43,10 +43,12 @@ main = do
   opts@Opts {..} <- execParser optsParserInfo
 
   cwd <- getCurrentDirectory
-  cfg <- case optInputFiles of
-    [] -> mkConfig cwd opts
-    ["-"] -> mkConfig cwd opts
-    file : _ -> mkConfig file opts
+  cfg <- case optConfigFilePath of 
+    Just configFilePath -> mkConfig configFilePath opts
+    Nothing -> case optInputFiles of
+      [] -> mkConfig cwd opts
+      ["-"] -> mkConfig cwd opts
+      file : _ -> mkConfig file opts
 
   let formatOne' =
         formatOne
@@ -272,6 +274,8 @@ data Opts = Opts
     optQuiet :: !Bool,
     -- | Ormolu 'Config'
     optConfig :: !(Config RegionIndices),
+    -- | Ormolu 'Config'
+    optConfigFilePath :: !(Maybe FilePath),
     -- | Fourmolu 'PrinterOpts',
     optPrinterOpts :: PrinterOptsPartial,
     -- | Options related to info extracted from files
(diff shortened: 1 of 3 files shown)


app/Main.hs
```diff
@@ -272,6 +274,8 @@ data Opts = Opts
     optQuiet :: !Bool,
     -- | Ormolu 'Config'
     optConfig :: !(Config RegionIndices),
+    -- | Ormolu 'Config'
```
brandonchinn178, line 277:
```suggestion
    -- | Path to Fourmolu config file
```

app/Main.hs
```diff
@@ -361,6 +365,11 @@ optsParser =
         help "Make output quieter"
       ]
     <*> configParser
+    <*> (optional . strOption . mconcat)
+      [ metavar "CONFIG_FILE",
+        long "config-file",
```
brandonchinn178, line 370: Any reason why not just `--config`? I think that's more standard with other CLI tools

app/Main.hs
```diff
@@ -43,10 +43,12 @@ main = do
   opts@Opts {..} <- execParser optsParserInfo
 
   cwd <- getCurrentDirectory
-  cfg <- case optInputFiles of
-    [] -> mkConfig cwd opts
-    ["-"] -> mkConfig cwd opts
-    file : _ -> mkConfig file opts
+  cfg <- case optConfigFilePath of 
+    Just configFilePath -> mkConfig configFilePath opts
```
brandonchinn178, line 47: I don't think this works? Did you test this?

I'm assuming you want to be able to do `--config-file /other/path/to/other_config_name_fourmolu.yaml`, but `mkConfig` will still look for a file named `fourmolu.yaml` in the directory named `configFilePath`.

I think this PR requires two changes:
1. Break up `loadConfigFile` into `findConfigFile` that does the logic to search for `fourmolu.yaml` and `loadConfigFile`, which should just decode the config file
2. Change `mkConfig` to take in `ConfigFileLoadResult` and in `main`, get that result from either `--config-file` => `loadConfigFile`, or `findConfigFile (cwd || head optInputFiles)`

flake.nix
```diff
@@ -0,0 +1,84 @@
+{
```
brandonchinn178, line 1: Is there a particular reason why this is being added? None of the current Fourmolu maintainers use Nix, and this file isn't being used in CI, so it'll be very prone to bitrot.

teto: sry I forgot to remove it.

app/Main.hs
teto: done

src/Ormolu/Config.hs
```diff
@@ -267,19 +268,22 @@ emptyConfig =
...
   findFile dirs configFileName >>= \case
     Nothing -> return $ ConfigNotFound dirs
-    Just file ->
-      either (ConfigParseError file) (ConfigLoaded file)
-        <$> Yaml.decodeFileEither file
+    Just file -> loadConfigFile file
+
+-- | Read options from a config file, if found.
+-- Looks recursively in parent folders, then in 'XdgConfig',
+-- for a file named /fourmolu.yaml/.
```
brandonchinn178, line 282: I believe this "look recursively" part should be documented for `findConfigFile` instead

brandonchinn178:
Reimplementing it here: https://github.com/fourmolu/fourmolu/pull/415

PR #403, opened 2024-03-09
Add ability to group imports

### LINKED ISSUE #297: Steal stylish-haskell's import grouping feature
`stylish-haskell` has recently acquired a nice feature for [configurably grouping imports](https://github.com/haskell/stylish-haskell/pull/399). It might be nice to steal it.

- [x] Yes, this feature is not relevant to Ormolu (because it's all about configurability, which ormolu is against)

michivi:
This PR attempts to group module import lists according to a selected strategy.

At first, I tried to allow the configuration file to control in a fine grain way how to regroup imports. But as mentioned in #297, the PR got very large and complex (especially for parsing), so I settled with this version that predefines some strategies.

I've only created strategies that I've been using in various projects. I'm aware that there are more, but I'd rather be sure that this PR is going the way you'd envisioned. They all evolve around:

- Grouping imports by whether they're qualified import or not
- Grouping imports by whether they're referring to modules in the same package or external

So this PR defines the following strategies:

- `none`: no regrouping (the default behavior before this PR)
- `by-qualified`: splits the import list by whether the import are qualified or not
- `by-scope`: splits the import list by whether they're referring to module from the current package or external
- `by-scope-then-qualified`: first splits by whether the imports are qualified or not, then by whether they're internal or external
- `by-qualified-then-scope`: first splits by whether the imports are internal or external, then by whether they're qualified or not

Also, I guess the names for the strategies are not quite clear, so open to any suggestion :slightly_smiling_face: 

Resolves #297


src/Ormolu/Imports.hs
```diff
@@ -6,43 +6,89 @@
...
+          . M.toAscList
+          . M.fromListWith combineImports
+          . fmap (\x -> (importId x, g x))
+      )
+    . ( if preserveGroups
+          then map toList . groupBy' (\x y -> not $ separatedByBlank getLocA x y)
+          else pure
+      )
   where
+    regroup :: GroupingStrategy -> [(ImportId, LImportDecl GhcPs)] -> [NonEmpty (ImportId, LImportDecl GhcPs)]
```
brandonchinn178, line 67: Can we break out `regroup` as a separate function?

src/Ormolu/Imports.hs
```diff
@@ -6,43 +6,89 @@
...
+
+    qualifiedGroups :: [(ImportId, LImportDecl GhcPs)] -> [NonEmpty (ImportId, LImportDecl GhcPs)]
+    qualifiedGroups imports =
+      let (unqualified, qualified) = partition (importQualified . fst) imports
+       in mapMaybe nonEmpty [qualified, unqualified]
+
+    scopeGroups :: Set Cabal.ModuleName -> [(ImportId, LImportDecl GhcPs)] -> [NonEmpty (ImportId, LImportDecl GhcPs)]
+    scopeGroups mods imports =
+      let (inScope, outOfScope) = partition ((`S.member` mods) . ghcModuleNameToCabal . importIdName . fst) imports
+       in mapMaybe nonEmpty [outOfScope, inScope]
```
brandonchinn178, line 90: It seems like the only purpose of the `NonEmpty` is the `mapMaybe nonEmpty`. It seems like it would be simplified a bit if we just replace `mapMaybe nonEmpty` with `filter (not . null)`?

src/Ormolu/Imports.hs
```diff
@@ -6,43 +6,89 @@
...
+
+data GroupingOperation
+  = UnqualifiedThenQualified
+  | GeneralThenSpecific !(Set Cabal.ModuleName)
+
+data GroupingStrategy
+  = -- | Does nothing beside normalization
+    NoGroupingStrategy
+  | -- | Group all normalized imports by applying the provided grouping operation
+    ApplyGroupingOperations !(NonEmpty GroupingOperation)
```
brandonchinn178, line 45: Why not just `[GroupingOperation]`, where no strategy is the natural consequence of `[]`?

src/Ormolu.hs
```diff
@@ -221,7 +221,10 @@ refineConfig sourceType mcabalInfo mfixityOverrides mreexports rawConfig =
             ],
       cfgDependencies =
         Set.union (cfgDependencies rawConfig) depsFromCabal,
-      cfgSourceType = sourceType
+      cfgSourceType = sourceType,
+      cfgDefinedModules =
+        cfgDefinedModules rawConfig
+          <> maybe Set.empty (Set.fromList . CabalUtils.ciModules) mcabalInfo
```
brandonchinn178, line 227: What's the point of `cfgDefinedModules`? When would you ever have modules that aren't defined in the cabal file?

web/site/site.hs
```diff
@@ -132,7 +134,7 @@ getPageInfo = \case
 
 getOptionDemoWidget :: ConfigData.Option -> Maybe String
 getOptionDemoWidget option@ConfigData.Option {..}
-  | name `elem` ["fixities", "reexports"] = Nothing
+  | name `elem` ["fixities", "reexports", "defined-modules"] = Nothing
```
brandonchinn178, line 137: Thanks for the care to update the web files!

brandonchinn178 (commented):
My biggest concern here is the hardcoded strategies. If we add more strategy types, the number of strategy options will increase exponentially, to handle all combinations of strategies. Why not specify the configuration as a list?

Also, I've never encountered a codebase that groups qualified imports together. The rise of `ImportQualifiedPost` seems to show that most people want to keep qualified/unqualified imports for the same module look even more similar. What's the motivation behind grouping by qualified-ness?

brandonchinn178:
I think I would much rather do a glob-based match or even a simple prefix-based match like stylish-haskell than add the complexity of parsing modules from the cabal file. For the most part, a project would be under the same hierarchy anyway, so it should be easy to define a pattern to match all of a project's modules.

For now, what do you think about just allowing configuration like
```yaml
import-groups:
- '*'
- 'Test.*'
- 'MyProject.*'
```
where `*` is only allowed to be at the end? Then matching should just be a matter of `isPrefixOf`

michivi (commented):
> I think I would much rather do a glob-based match or even a simple prefix-based match like stylish-haskell than add the complexity of parsing modules from the cabal file. For the most part, a project would be under the same hierarchy anyway, so it should be easy to define a pattern to match all of a project's modules.

> For now, what do you think about just allowing configuration like

> import-groups:
> - '*'
> - 'Test.*'
> - 'MyProject.*'

> where * is only allowed to be at the end? Then matching should just be a matter of isPrefixOf

Not having to parse the cabal configuration would definitely make things simpler. I do have projects which are not under the same hierarchy root, but it's easy enough to just enumerate the project's (few) parent modules so that should be fine.

The reasons I went with those strategies was to keep things simple for now: I saw projects which grouped imports by whether they were qualified or not, by whether it's internal or not, by actual parent packages... And the order of these resulting groups also varies. But I have not seen statistics on them (which strategy is used the most, how much configurability would developers want...). So I thought I would just give it a shot with this PR with predefined strategies. My thinking was that refactoring to include additional strategies, or use the configuration strategy you've defined would be easier to implement with this base.

The only thing I'm wondering about with your suggestion configuration is: how would you handle qualified imports? If qualified imports should be handled just like any other imports, then I guess it'll be straightforward. But if projects want to isolate those qualified imports, then some additional configuration would be required. So I guess: should fourmolu allow developers to customize the grouping based on whether the imports are qualified? If so, how much configurability should it give to the developers?

src/Ormolu/Imports.hs
brandonchinn178: Sure, but we're not actually using the invariant. I'd recommend seeing what the code looks like with `[[(ImportId, LImportDecl GhcPs)]]`, and if it's better change it, if not, keep as NonEmpty

src/Ormolu/Imports.hs
brandonchinn178: I think `[]` will be clearer. As programmers, I think we're all comfortable with the idea of "do no actions" being the same as "noop" 🙂

brandonchinn178:
> So I guess: should fourmolu allow developers to customize the grouping based on whether the imports are qualified? If so, how much configurability should it give to the developers?

I just noticed that @georgefst commented about separating imports in his project by qualified, which I missed earlier. It's interesting, because I've never seen a project do that before!

IMO doing a regex-based style will handle 80% of use cases from people who've been asking for this feature. I think we should just do that for now, and then we can see what the next biggest import strategy people want after.

src/Ormolu/Imports.hs
michivi: Done. In the end, it was simpler to integrate with the existing code that did not make use of `NonEmpty`.

src/Ormolu/Imports.hs
michivi: Done!

michivi:
I'll replace the "defined-modules" part through Cabal file parsing with a regular expressions. I'll also try to see if we can handle qualified in an easy and configurable way without the current "strategies" approach in this PR. Converting back to draft in the meantime.

michivi:
I'm testing a new configurable implementation. The code needs some cleaning up, but it is at least operational. It supports the following configurations:

```yaml
# Just use a single group (default)
import-grouping: single

# Regroups by whether the imports are qualified / defined by the current project / both.
# Those presets are for easy configuration and can be changed.
import-grouping: by-qualified
import-grouping: by-scope
import-grouping: by-scope-and-qualified

# A more complex configuration for users who know what they want (... with a silly example)
import-grouping:
  - name: "Text modules"
    rules: # A module must satisfy at least one of those rule to be in this group.
      - regex: "Data\\.Text(\\..+)?" # Matches modules by regex
  - name: "The rest"
    rules:
      - match: all # Match every module
        tie-breaker: 255 # If a module is matched by multiple rules from different group, the lowest tie-breaker will win.
  - name: "My internals and monads unqualified"
    rules:
      - cabal: defined-modules # Matches all modules defined within the current Cabal project. Will be removed in favor or regex.
      - regex: "Control\\.Monad(\\..+)?"
  - name: "My internals and monads qualified"
    rules:
      - cabal: defined-modules
        qualified: yes # Can restrict on whether the import is qualified or not (by default, matches both)
      - regex: "Control\\.Monad(\\..+)?"
        qualified: yes
  - name: "Monad State"
    rules:
      - regex: "Control\\.Monad\\.State\\.Lazy"
        tie-breaker: 1
```

I've kept the "Fetching the Cabal project's defined modules" for now, but we can remove it later. It's proving useful for easy  testing.

This configuration schema was designed to be as easy to use as possible (i.e., usable presets, explicit behavior, not too nested). It does introduce some complexity on the implementation side. And I haven't even started adding the YAML configuration panel for the site (if we go through with this, it will deserve an entire PR on its own... not even sure how it would look like besides a YAML editor).

Is this going in a good direction?

brandonchinn178:
Not at my laptop right now, so I can't review just yet. But overall, regex seems a bit too powerful for this feature. Plus, the backslashes are annoying in YAML. What about using globs instead? I think globs would satisfy the majority of use cases

brandonchinn178:
Also, instead of tie-breaker (which seems a bit difficult to understand), what about defining the rules (top-down priority) separately from the group order? e.g.

```yaml
import-groups:
  rules:
    - name: text
      match:
        - pattern: Data.Text.* # also matches Data.Text
    - name: state
      match:
        - pattern: Control.Monad.State.*
    - name: internals-and-monads-unqualified
      match:
        - builtin: defined-modules
        - pattern: Control.Monad.*
    - name: rest
      match:
        - builtin: all
  groups:
    - text
    - rest
    - internals-and-monads-unqualified
    - state
```

michivi:
> But overall, regex seems a bit too powerful for this feature. Plus, the backslashes are annoying in YAML. What about using globs instead? I think globs would satisfy the majority of use cases

Thinking about it, I'm even wondering if globs would be too powerful. I'm usually only interested in grouping specific modules and their descendants.

> Also, instead of tie-breaker (which seems a bit difficult to understand), what about defining the rules (top-down priority) separately from the group order?

In hindsight, `tie-breaker` is indeed not that clear. I've renamed it to `priority`. But I think having a separate section for the order is too verbose and somewhat redundant (one section for defining the rules, another one to define the order of application).

Thinking about it, the priority would only required when there's conflict in rules (a module is eligible for multiple groups). But if we're only considering the module's prefix most of the time, then it should be rare. It mostly happens for general "catch-all" rules. But those can be defined as preset with the lowest priority. Plus, they're usually the same throughout projects (all qualified, all non qualified...). So we might get away with the following middle ground:


```yaml
import-grouping:
  - name: "Everything else"
    preset: all # lowest priority
  - name: "Our modules"
    preset: defined-modules # almost the lowest priority
  - name: "Modules we want emphasize for some reasons"
    rules:
      - module-or-descendant: Custom.Very.Important.Module
        # the default priority is enough to win over the presets
```

and the more complex and silly example:

```yaml
import-grouping:
  - name: "Text modules"
    rules:
      - module-or-descendant: Data.Text
  - name: "The rest"
    preset: all
  - name: "My internals and monads unqualified"
    rules:
      - cabal: defined-modules
        qualified: no
      - regex: "Control\\.Monad(\\..+)?"
        qualified: no
  - name: "My internals and monads qualified"
    rules:
      - cabal: defined-modules
        qualified: yes
      - regex: "Control\\.Monad(\\..+)?"
        qualified: yes
  - name: "Monad State"
    rules:
      - module-or-descendant: "Control.Monad.State.Lazy"
        priority: 1 # only because it conflicts with our previous custom rule
```

That way, the configuration only needs to be as complex as the rules we want to express.

What do you think?

brandonchinn178:
I'm using "glob" fairly loosely. I was thinking it could be as simple as "star is only allowed at the end", so a "glob" is effectively either an exact match or a prefix match. But co-opting glob syntax, I think, is a bit easier to read than `module-or-descendent`. And it leaves the possibility open for full glob support, if it comes up.

It looks like your complicated silly example still has regex? Can we just drop regex support altogether? It would be great to avoid adding the regex dependency 😄 

In general, I would be hesitant to make the group-order and the priority-order the same, because a user might run into an issue where they want to flip the order of two groups but can't because the priority of the rule is coupled to the order of the groups. But one possibility is to define priority similar to CSS: a given module matches the most specific pattern, regardless what order the patterns are in.

michivi:
> I'm using "glob" fairly loosely. I was thinking it could be as simple as "star is only allowed at the end", so a "glob" is effectively either an exact match or a prefix match. But co-opting glob syntax, I think, is a bit easier to read than `module-or-descendent`. And it leaves the possibility open for full glob support, if it comes up.

Agreed. I've included a basic implementation for globbing (which work on any strings, not necessarily modules so it doesn't consider `.` as delimiters) and remove `module-or-descendent`.

> It looks like your complicated silly example still has regex? Can we just drop regex support altogether? It would be great to avoid adding the regex dependency 😄

Sure, the current draft is no longer using it. :slightly_smiling_face: 

> In general, I would be hesitant to make the group-order and the priority-order the same, because a user might run into an issue where they want to flip the order of two groups but can't because the priority of the rule is coupled to the order of the groups. But one possibility is to define priority similar to CSS: a given module matches the most specific pattern, regardless what order the patterns are in.

I see your point.

CSS-like priority (through specificity) is interesting, but hard to pull off in our case as 1) we're not necessarily making use of globs (we currently have the `defined-modules` matcher, but even if we remove it, shouldn't we stay open to additional matcher?) and 2) how would you define the specificity of a glob? By length seems arbitrary... By depth means the glob needs to be aware of the module path. But then, it would still apply priority to a single rule, not to the group as you've pointed out.

I'm wondering if this priority issue, while important, may not be urgent. Should we keep the notion of priority implicit and private for now (so that preset are still of a lesser priority than custom rules) until we have more feedback on the import grouping feature? Or is it too important and impacting, so needs to be figure out as soon as possible?

brandonchinn178:
Very cool, this is shaping up to be quite a feature!

I agree, let's leave priority implicit for now, and we can always surface it later.

One minor nit: it bugs me that `Control.Monad*` could match `Control.MonadFoo` also. Personally, I would prefer the glob be module-aware, where `Control.Monad.*` represents `Control.Monad` and `Control.Monad.Foo`, but not `Control.MonadFoo`. Thoughts?

michivi:
> One minor nit: it bugs me that `Control.Monad*` could match `Control.MonadFoo` also. Personally, I would prefer the glob be module-aware, where `Control.Monad.*` represents `Control.Monad` and `Control.Monad.Foo`, but not `Control.MonadFoo`. Thoughts?

Agreed.

At first, I thought it would be simpler to ignore module levels and consider names as just strings. But I just realized that `Control.Monad.*.Lazy` would not only match `Control.Monad.Writer.Lazy` but also `Control.Monad.Imaginary.Monad.Lazy` as well, which may not be what we want, if we only want `*` to match a single level.

Let me try to implement `*` as "anything but on the same level", and `**` as "anything, but that can span 0 to multiple level", like regular filesystem glob.

michivi:
Implemented, though `Control.Monad.**` currently won't match `Control.Monad`. But wondering if it would be a good idea, as I have arguments for both scenarios. And it's pretty easy to work around (one rule for `Control.Monad` and one for `Control.Monad.**` in the same group).

There's also the question of whether the introduced configuration complexity is acceptable. We should be fine wrt to Ormolu as the complexity is isolated, but it is still quite different from the existing configuration.

What do you think?

brandonchinn178:
I think this is fine for now. So * matches anything except dot and ** matches anything including dot?

michivi:
> I think this is fine for now. So * matches anything except dot and ** matches anything including dot?

Right! And there's also the question of whether the `cabal: defined-modules` should stay or be removed (or be refactored).

brandonchinn178:
I'll leave cfgDefinedModules up to you for now. Just mark the PR ready for review when you're ready and I can review it then

michivi:
Ok, I've left `defined-modules` as it's used in existing presets. I'm open to removing it ,but while simplification is a real plus, we'd be left with a single preset :sweat_smile:

config/FourmoluConfig/ConfigData.hs
```diff
@@ -228,6 +228,26 @@ allOptions =
...
+        fieldName = Just "poImportGrouping",
+        description = "Rules for grouping import declarations",
+        type_ = "ImportGrouping",
+        default_ = HsExpr "CreateSingleGroup",
+        ormolu = HsExpr "CreateSingleGroup",
+        sinceVersion = Nothing,
+        cliOverrides = emptyOverrides
+      },
+    Option
+      { name = "defined-modules",
```
brandonchinn178, line 243: `defined-modules` doesn't seem very clear; what does it mean for a module to be defined? Perhaps `local-modules`?

Same for `MatchDefinedModules`

web/site/pages/config/defined-modules.md
```diff
@@ -0,0 +1,7 @@
+# `defined-modules`
+
+$info$
+
+You usually won't need to specify those manually. Fourmolu will fetch those from the Cabal project file.
+
+See [`import-grouping`](/config/import-grouping) for more info.
```
brandonchinn178, line 7:
```suggestion
Modules to consider as part of the current Cabal package, for the [`import-grouping`](/config/import-grouping) option.

This is automatically detected from the Cabal file, but can be useful to set manually when you don't have a Cabal file.
```

web/site/pages/config/import-grouping.md
```diff
@@ -0,0 +1,147 @@
...
+- `by-scope`: one group for external imports, then one group for imports targeting modules from the current Cabal project
+- `by-scope-then-qualified`: apply `by-scope` first, then `by-qualified`
+
+Predefined groups:
+
+- `all`: preset group that will receive any import not matching any of your custom rules
+
+Predefined matchers:
+
+- `cabal: defined-modules`: matches modules defined in the current Cabal project
```
brandonchinn178, line 25: can you mention that this is automatically detected from the cabal file, but modules can be included with the `--defined-modules` CLI flag?

web/site/pages/config/import-grouping.md
```diff
@@ -0,0 +1,147 @@
...
+by-scope
+{ "import-grouping": "by-scope", "respectful": false }
+```
+
+```fourmolu-example-tab
+by-scope-then-qualified
+{ "import-grouping": "by-scope-then-qualified", "respectful": false}
+```
+
+```fourmolu-example-tab
```
brandonchinn178, line 92: Can you add a markdown comment here saying to keep in sync with the configuration above?

web/site/pages/config/import-grouping.md
```diff
@@ -0,0 +1,147 @@
...
+Predefined groups:
+
+- `all`: preset group that will receive any import not matching any of your custom rules
+
+Predefined matchers:
+
+- `cabal: defined-modules`: matches modules defined in the current Cabal project
+- `glob: pattern`: matches modules matching the provided `pattern`. `*` can be any character on the same module level. `**` can be any character and can span multiple module levels.
+
+Here's a sample configuration:
```
brandonchinn178, line 28: Can you mention that this is the configuration used in the "custom" example?

src/Ormolu/Imports.hs
```diff
@@ -6,39 +6,214 @@
...
+
+    convertGroupPreset :: Config.ImportGroupPreset -> NonEmpty ImportGroupRule
+    convertGroupPreset = \case
+      Config.AllPreset ->
+        NEL.singleton
+          ImportGroupRule
+            { igrModuleMatcher = MatchAllModules,
+              igrQualifiedMatcher = MatchBothQualifiedAndUnqualified,
+              igrPriority = maxBound
+            }
```
brandonchinn178, line 140: What's the point of `AllPreset`? Why would you ever use `preset: all` when you can do `match: all`?

Furthermore, what's the point of `match: all` when you can do `glob: **`?

src/Ormolu/Imports.hs
```diff
@@ -6,39 +6,214 @@
...
+      igrQualifiedMatcher = MatchBothQualifiedAndUnqualified,
+      igrPriority = maxBound
+    }
+
+matchModulesRule :: Set Cabal.ModuleName -> ImportGroupRule
+matchModulesRule mods =
+  ImportGroupRule
+    { igrModuleMatcher = MatchModules mods,
+      igrQualifiedMatcher = MatchBothQualifiedAndUnqualified,
+      igrPriority = ImportRulePriority 192 -- Lower priority than "all" but higher than the default.
```
brandonchinn178, line 171: These numbers are very magical 😄 Can we just do 
* default = 50
* match modules = 60
* all = 100
?

src/Ormolu/Imports.hs
```diff
@@ -6,39 +6,214 @@
...
 import GHC.Types.SourceText
 import GHC.Types.SrcLoc
-import Ormolu.Utils (groupBy', notImplemented, separatedByBlank, showOutputable)
+import Ormolu.Config qualified as Config
+import Ormolu.Utils (ghcModuleNameToCabal, groupBy', notImplemented, separatedByBlank, showOutputable)
+import Ormolu.Utils.Glob (Glob, matchesGlob, mkGlob)
+
+newtype ImportGroups = ImportGroups (NonEmpty ImportGroup)
+
+data ImportGroup = ImportGroup
```
brandonchinn178, line 42: IIUC, the difference between this and `Config.ImportGroup` is that the config version is the "static specification for an import group" and this is the "import group rule that's desugared and ready to execute", right?

I don't see a meaningful difference between the two use-cases. Why do we need an intermediate data type here? Why not desugar directly to these rules in FromJSON? The only hurdle is `MatchModules (Set ModuleName)`, but we could keep `MatchDefinedModules` and just explicitly pass in `definedModules` to `normalizeImports`

brandonchinn178 (commented):
Sorry for the late review, I didn't get a notification that you marked the PR ready for review.

Also sorry for the deluge of comments. Major design questions (feel free to respond in the relevant in-line comments):
1. Why have two `ImportGroups` types? Why not just parse to the desugared AST?
2. Why not define `FromJSON` instances in `Ormolu.Config.Fixed`?
3. What do you think about renaming `defined-modules` to `local-modules`?

src/Ormolu/Imports.hs
michivi: They are! 🙂 I just wanted some powers of 2... for some reasons 😅
We can go with your values.

src/Ormolu/Imports.hs
```diff
@@ -52,6 +227,26 @@ normalizeImports preserveGroups =
             ..
           }
 
+groupImports :: ImportGroups -> [(ImportId, x)] -> [[(ImportId, x)]]
```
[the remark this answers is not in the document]
michivi: It would be nice, but it would introduce a dependency cycle (`Ormolu.Imports` would import `groupImports` from `Ormolu.Imports.Grouping`, which would import `ImportId` from `Ormolu.Imports`). It would be a good idea to move `ImportId` out to prevent this, but then it would be quite a diverge from `ormolu`. What do you think? Perhaps at the very least some data types can be moved out...

src/Ormolu/Config/Fixed.hs
```diff
@@ -0,0 +1,36 @@
...
+
+data ImportGroupPreset
+  = -- | Preset that will group every import declarations not matched by custom rules
+    AllPreset
+  deriving (Eq, Show)
+
+data ImportGroupRule = ImportGroupRule
+  { igrModuleMatcher :: !ImportModuleMatcher,
+    -- | 'Just True' to match qualified declarations, 'Just False' to match unqualified ones and 'Nothing' to match both
+    igrQualified :: !(Maybe Bool)
```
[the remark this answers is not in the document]
michivi: For the reasons mentioned [here](https://github.com/fourmolu/fourmolu/pull/403#discussion_r1623265634)

src/Ormolu/Config/Fixed.hs
```diff
@@ -0,0 +1,36 @@
...
+  }
+  deriving (Eq, Show)
+
+data ImportModuleMatcher
+  = MatchAllModules
+  | MatchDefinedModules
+  | MatchGlob !String
```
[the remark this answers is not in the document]
michivi: For the reasons mentioned [here](https://github.com/fourmolu/fourmolu/pull/403#discussion_r1623265634)

src/Ormolu/Config/Fixed.hs
```diff
@@ -0,0 +1,36 @@
...
+data ImportGroup = ImportGroup
+  { igName :: !(Maybe String),
+    igPresetOrRules :: !(Either ImportGroupPreset (NonEmpty ImportGroupRule))
+  }
+  deriving (Eq, Show)
+
+data ImportGroupPreset
+  = -- | Preset that will group every import declarations not matched by custom rules
+    AllPreset
+  deriving (Eq, Show)
```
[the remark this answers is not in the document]
michivi: For the reasons mentioned [here](https://github.com/fourmolu/fourmolu/pull/403#discussion_r1623265634)
(sorry, some questions actually boil down to the same answer :sweat_smile: )

michivi:
I'll try to address the remaining points as soon as possible, but it may take some time.

src/Ormolu/Config/Types.hs
```diff
@@ -0,0 +1,93 @@
+{-# LANGUAGE LambdaCase #-}
+{-# LANGUAGE OverloadedStrings #-}
+
+-- | This module contains configuration data types that are not meant to be
+-- provided directly as CLI arguments or through the configuration GUI but as
+-- more advanced YAML options.
```
brandonchinn178, line 6: I think we can just remove this comment?

src/Ormolu/Config/Types.hs
```diff
@@ -0,0 +1,93 @@
...
+        parseMatchModuleMatcher v,
+        parseGlobModuleMatcher v
+      ]
+    where
+      parseCabalModuleMatcher :: Aeson.Value -> Aeson.Parser ImportModuleMatcher
+      parseCabalModuleMatcher = Aeson.withObject "ImportModuleMatcher" $ \o -> do
+        c <- Aeson.parseField @String o "cabal"
+        case c of
+          "local-modules" -> pure MatchLocalModules
+          other -> Aeson.parseFail $ "Unknown Cabal matching: " <> other
```
brandonchinn178, line 81: Can't this be `match: local-modules`?

tests/Ormolu/Config/PrinterOptsSpec.hs
```diff
@@ -247,17 +247,70 @@ spec =
...
           isMulti = False,
           testCases =
-            let testedStrategies = [CreateSingleGroup, SplitByQualified, SplitByScope, SplitByScopeAndQualified]
+            let testedStrategies =
+                  [ CreateSingleGroup,
+                    SplitByQualified,
+                    SplitByScope,
+                    SplitByScopeAndQualified,
+                    UseCustomImportGroups $
+                      NonEmpty.fromList
```
brandonchinn178, line 256: Completely optional tip: One thing I like to do to reduce indentation here is
```hs
UseCustomImportGroups . NonEmpty.fromList $
  [ ...
  ]
```

tests/Ormolu/ConfigSpec.hs
```diff
@@ -31,3 +32,274 @@ spec = do
...
+        config <- Yaml.decodeThrow "import-grouping: by-scope-then-qualified"
+        let actualStrategy = poImportGrouping (cfgFilePrinterOpts config)
+        actualStrategy `shouldBe` Just SplitByScopeAndQualified
+      it "fails when an unknown strategy is requested" $ do
+        let parse :: IO FourmoluConfig
+            parse = Yaml.decodeThrow "import-grouping: fake-strategy"
+            anUnknownStrategyValue e = case e of
+              Yaml.AesonException msg -> "unknown strategy value" `isInfixOf` msg
+              _ -> False
+        parse `shouldThrow` anUnknownStrategyValue
```
brandonchinn178, line 59: how about
```hs
Left (Yaml.AesonException msg) <- Yaml.decodeEither' "import-grouping: fake-strategy"
msg `shouldSatisfy` ("unknown strategy value" `isInfixOf`)
```

tests/Ormolu/Config/PrinterOptsSpec.hs
```diff
@@ -241,6 +242,23 @@ spec =
           testCaseSuffix = \(respectful, importExportStyle) ->
             suffixWith ["respectful=" ++ show respectful, show importExportStyle],
           checkIdempotence = True
+        },
+      TestGroup
+        { label = "import-grouping",
+          isMulti = False,
+          testCases =
+            let testedStrategies = [CreateSingleGroup, SplitByQualified, SplitByScope, SplitByScopeAndQualified]
```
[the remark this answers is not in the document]
brandonchinn178: IMO if you're specifying `import-grouping`, you probably want imports to be completely automatically formatted, overriding any idiosyncratic groupings. e.g. HLS always adds imports to the bottom of the last group, and with this option specified, the user probably expects fourmolu to reformat everything.

Add for context, `respectful` should generally always be true. false is mostly for Ormolu backwards-compatibility. So I think we can completely ignore respectful here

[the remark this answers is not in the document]
brandonchinn178: Or going further, I think we should remove "import group behavior" from respectful completely, and change the default `import-grouping: null` to "leave groups unchanged" (the current `respectful: true`) and use `import-grouping: single` to get back `respectful: false`

src/Ormolu/Imports.hs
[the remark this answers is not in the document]
brandonchinn178: I think breaking out the 200 lines here is worth a little duplication. I think we can define `Ormolu.Imports.Grouping.ImportId` with just `importIdName` and `importQualified`, and then you can do one of two things:
1. Convert back and forth in `normalizeImports`
2. Define `groupImports` as
    ```hs
    (importId -> ImportId)
    -> ImportGroups
    -> [(importId, x)]
    -> [(importId, x)]
    ```
    Then `normalizeImports` would pass the conversion function in

src/Ormolu/Utils/Glob.hs
```diff
@@ -0,0 +1,41 @@
...
+        null s
+    MatchExactly p g' ->
+        case stripPrefix p s of
+            Nothing -> False
+            Just s' -> matchesGlob s' g'
+    SingleWildcard g' ->
+        let l = fromMaybe (length s) (elemIndex '.' s)
+         in or [matchesGlob s' g' | i <- [0 .. l], let s' = drop i s]
+    DoubleWildcard g' ->
+        or [matchesGlob s' g' | i <- [0 .. length s], let s' = drop i s]
```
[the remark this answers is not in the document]
brandonchinn178: No worries. I think there are some other examples in the new commits you just pushed. Can you rebase against main? I just added a pre-commit check that'll fix these newlines. You can just run `pre-commit run -a` to fix them

src/Ormolu/Utils/Glob.hs
```diff
@@ -0,0 +1,41 @@
...
+    ( Glob,
+      mkGlob,
+      matchesGlob,
+    )
+where
+
+import Data.List (elemIndex, stripPrefix)
+import Data.Maybe (fromMaybe)
+
+data Glob
```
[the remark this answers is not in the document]
brandonchinn178: I would prefer the list, mostly to get rid of `GlobCase` in the test file. See https://github.com/fourmolu/fourmolu/pull/403#discussion_r1623164652

If it doesnt simplify the test, then this is fine as-is

config/FourmoluConfig/ConfigData.hs
```diff
@@ -430,5 +450,62 @@ allFieldTypes =
...
           ]
+      },
+    FieldTypeADT
+      { fieldTypeName = "ImportGrouping",
+        adtConstructors =
+          [ "CreateSingleGroup",
+            "SplitByScope",
+            "SplitByQualified",
+            "SplitByScopeAndQualified",
+            "UseCustomImportGroups (NonEmpty CT.ImportGroup)"
```
brandonchinn178, line 461: Nit: Can we rename to
```suggestion
          [ "ImportGroupSingle",
            "ImportGroupByScope",
            "ImportGroupByQualified",
            "ImportGroupByScopeThenQualified",
            "ImportGroupCustom (NonEmpty CT.ImportGroup)"
```
Name should be declarative, not imperative. Names should also be consistent, and it should be easy to identify which option it belongs to, just from the constructor name

src/Ormolu/Imports.hs
brandonchinn178: It seems to me like any preset we add could be added to `match`? Like we could define `match: monads` as an alias for
```yaml
- oneof:
  - glob: Control.Monad
  - glob: Control.Monad.**
```
I vote getting rid of presets and just using `match`

src/Ormolu/Imports.hs
brandonchinn178: Ha I appreciate your reasoning here, I argued for exactly this at my work: https://brandonchinn178.github.io/posts/2023/04/15/sharing-data-types. So in principle, I completely agree with you.

Both of your points relate to runtime datatypes being lossy, and keeping runtime and config separate allows for recovery of the exact user input. It's true that right now, we don't need to recover the exact user input, but furthermore, it's also true that the current runtime `ImportGroup` type is lossless. Even if we needed to return the user config again, we could recover it directly from the runtime type. We could also do the same if we added regex, with a new `Regex` type that stored the user input with the parsed pattern.

Plus, let's say we add regex, or add the notion of an invalid glob. We'd want to do the parsing (and failing) in the `parseJSON` step, right? And we don't want to reparse afterwards, so the config type _should_ contain the parsed type.

The only difference left is `ImportRulePriority`, but I'd argue that can also go in the config, as that's something that we might reasonably support in configuration somehow anyway.

So I don't see any current motivation to separate runtime from config, and if a reason comes up later, we can always re-split.

---

As a quick sketch (consolidating my other comments), I'm thinking
```hs
-- generated
data ImportGrouping
  = ImportGroupSingle
  | ImportGroupByScope
  | ImportGroupByQualified
  | ImportGroupByScopeThenQualified
  | ImportGroupCustom (NonEmpty ImportGroup)

data ImportGroup = ImportGroup
  { groupName :: Maybe String
  , groupRules :: NonEmpty ImportGroupRule
  }

data ImportGroupRule = ImportGroupRule
  { ruleMatch :: ImportGroupRuleMatch
  , ruleQualified :: ImportGroupRuleQualified
  , rulePriority :: Int
    -- ^ currently hardcoded for each RuleMatch, but
    -- might be configurable in future
  }

data ImportGroupRuleQualified
  = MatchQualified
  | MatchUnqualified
  | MatchAnyQualified

data ImportGroupRuleMatch
  = MatchAll
  | MatchLocalModules
  | MatchGlob Glob
```

src/Ormolu/Imports.hs
michivi: Done. Though what bothers me now is that we actually have:

```yaml
preset: all
```

replaced by

```yaml
rules:
  - match: all
    priority: 100
```

So we need add `priority` in the configuration. That's what I did for now.

src/Ormolu/Imports.hs
michivi: So I've updated the PR. While at first it seemed that it simplifies some code, I feel some other are actually less simple. For example: I agree that we only need to parse Glob once during parsing... but now tests have to create globs to test unrelated things. To be fair, we could just test fields we are interested in, but we'd need accessors from `ImportGrouping` for each field we want to test. Still trying to figure out how to do that in a simple and maintainable way.

Also, Trying to see how to handle the "local modules" part, since we don't have access to those during configuration parsing. That's this PR, in its current state, still has some duplication to handle this case.

src/Ormolu/Config/Types.hs
michivi: Done

src/Ormolu/Config/Types.hs
michivi: Right. It does simplify things. Done.

tests/Ormolu/Config/PrinterOptsSpec.hs
michivi: Done

tests/Ormolu/ConfigSpec.hs
michivi: Made something similar as this exact syntax fails, and it feels weird lifting into `IO`...

config/FourmoluConfig/ConfigData.hs
michivi: Right, done!

tests/Ormolu/Config/PrinterOptsSpec.hs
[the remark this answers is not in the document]
michivi: Agreed. I'll tackle that after the other feedback points have settled, as I suspect it will required some refactoring :slightly_smiling_face:

src/Ormolu/Imports.hs
michivi: Oh, just realized that using a single data type group would also result in the loss of "split by scope" (and so on) information, as they are not used at runtime. So it's not lossless as you've mentioned. But it also may not be that problematic: we could just have the configuration parsing code transforming that into actual rules. I'm not a fan of that solution, but exact lossless parsing is still not an absolute requirement at the moment...

brandonchinn178: > but now tests have to create globs to test unrelated things

I mean, the test does create the glob at some point, right? It's just a question of whether you create the glob in the test, or the function being tested is parsing the string into a glob. If the test doesn't care about the glob, then just define an opaque `testGlob :: Glob` helper that a test can just use as an arbitrary glob

> Oh, just realized that using a single data type group would also result in the loss of "split by scope" (and so on) information, as they are not used at runtime.

What do you mean? After parsing the config, you now have `ImportGroupByScope`, and then you can just do
```hs
resolveGroups = \case
  ImportGroupSingle -> groupsSingle
  ImportGroupByScope -> groupsByScope
  ...
  ImportGroupCustom groups -> groups
```

> Also, Trying to see how to handle the "local modules" part, since we don't have access to those during configuration parsing.

My suggestion was to leave it as `ByLocalModules` and then just explicitly pass in `definedModules` to `normalizeImports`

tests/Ormolu/Utils/GlobSpec.hs
```diff
@@ -0,0 +1,67 @@
...
+            "MyApp42.Control.Monad" `matchesGlob` mkGlob "*42.Control.Monad" `shouldBe` True
+        it "should allow * at the end" $
+            "System.IO.Stuff" `matchesGlob` mkGlob "System.IO.*" `shouldBe` True
+        it "should allow ** to match anything on the same level" $
+            "Level1.Level2.Level3" `matchesGlob` mkGlob "Level1.L**2.Level3" `shouldBe` True
+        it "should allow ** to match anything on multiple levels" $
+            "Level1.Level2.Level3.Level4.Level5" `matchesGlob` mkGlob "Level1.L**4.Level5" `shouldBe` True
+        it "should match strings where wildcards can be replaced by 0 or more characters" $
+            property $
+                \gc -> gcMatchingString gc `matchesGlob` gcGlob gc
```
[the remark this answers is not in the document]
brandonchinn178: Bump, in case you missed this

tests/Ormolu/Utils/GlobSpec.hs
```diff
@@ -0,0 +1,67 @@
...
+import Test.QuickCheck (Arbitrary (..), listOf, oneof, property, suchThat)
+
+data GlobCasePart
+    = SingleWildcardFor !String
+    | DoubleWildcardFor !String
+    | ExactPart !String
+    deriving (Eq, Show)
+
+newtype GlobCase = GlobCase [GlobCasePart]
+    deriving (Eq, Show)
```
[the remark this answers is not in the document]
brandonchinn178: Bump in case you missed this

tests/Ormolu/ConfigSpec.hs
brandonchinn178: What was the failure? And what do you mean by lifting into `IO`? My suggestion does the same thing the other tests do, by running `Yaml.decodeThrow` directly in the `Expectation`

src/Ormolu/Imports.hs
brandonchinn178: Why do you need to add `priority`? Why not just explicitly set the priority field to 100 when parsing the `match: all` rule?

tests/Ormolu/Config/PrinterOptsSpec.hs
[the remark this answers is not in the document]
michivi: Added `import-grouping: legacy` to simplify the code that represents the backward compatible behavior. This behavior being dependent from `respectful`, I'd prefer to keep the parsing code independent, and have the import code adapting based on `import-grouping` and `respectful`.

src/Ormolu/Imports.hs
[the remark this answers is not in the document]
michivi: Addressed. There are still tests missing, I'll come back to this when I have some time later.

src/Ormolu/Utils/Glob.hs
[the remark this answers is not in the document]
michivi: Addressed using list only.

tests/Ormolu/Utils/GlobSpec.hs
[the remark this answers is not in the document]
michivi: I've simplified tests. I still find property tests somewhat lacking. If you have ideas for other properties :slightly_smiling_face:

web/site/pages/config/defined-modules.md
michivi: Addressed.

web/site/pages/config/import-grouping.md
michivi: Addressed. Is the documentation clear enough on that point?

web/site/pages/config/import-grouping.md
michivi: I did, though as an HTML comment (I don't think the Markdown implementation we're using handles comment, does it?). As a consequence, it will appear in the generated web site (in the source code).

web/site/pages/config/import-grouping.md
michivi: Addressed.

src/Ormolu/Imports.hs
michivi: I hesitated doing that, but while the `100` priority would be explicit in the application's code, it won't be in the configuration. I tried to avoid implicit configuration (e.g., when using `match: all` your rule will have a priority of `100`, otherwise, it will be `50`). Does that make sense?

tests/Ormolu/ConfigSpec.hs
michivi: We have `decodeEither' :: FromJSON a => ByteString -> Either ParseException a`. There's no base monad for us to match `Left _` in. So we'll end up with:
```
    • Couldn't match type ‘IO’ with ‘Either Yaml.ParseException’
      Expected: Either Yaml.ParseException ()
        Actual: Expectation
```

Lifting in `IO` with `pure` before `Yaml.decodeEither'` can solve this but it doesn't feel satisfactory. That's why I went with `decodeThrow`. Does that make sense? Am I missing something?

michivi:
@brandonchinn178 I believe I've addressed all your feedback (hopefully, I haven't missed anything :sweat_smile: ). Sorry for the time delay.

The only thing left that is bugging me is the `priority` configuration, as mentioned in my [reply](https://github.com/fourmolu/fourmolu/pull/403#discussion_r1639830468). We could force priority to be `100` for `match: all`, but I find that implicit configuration changing (in this case `priority` depending on the `match` rule) can be disturbing and prone to errors (discrepancies with what the user expects). That being said, going with your proposal, we could argue that if we hide `priority` configuration key in this version, then there's no implicit configuration, just an implicit behavior change (i.e., the `match: all` rule is of a less priority than the rest). This doesn't remove the expectation issue. What do you think? Would you still go with an implicit "less priority rule"?

web/site/pages/config/import-grouping.md
brandonchinn178: Yes, "markdown comment" = "HTML comment". It's fine if it shows up in the source code

src/Ormolu/Imports.hs
brandonchinn178: I mean, in your original implementation, the priority was implicit too, right? In the new implementation, all that's changed is the implicit priority is in the same data type the configuration parses to. So I don't really see a difference, except now we can reuse the type instead of having two `ImportGroup` types. And if we want to expose the priority in configuration, it's easier now -- you just parse the priority field instead of inferring it

tests/Ormolu/ConfigSpec.hs
brandonchinn178: Ah yes, that's correct. My suggestion should be `pure $ Yaml.decodeEither' ...`. I do still think it's a bit better, as there's no point throwing an exception just to catch it and assert on it.

If you really want to avoid the `pure`, perhaps
```hs
let res = Yaml.decodeEither' "import-grouping: fake-strategy"
res `shouldSatisfy` \case
  Left e -> "unknown strategy value" `isInfixOf` e
  Right _ -> False
```

tests/Ormolu/Utils/GlobSpec.hs
```diff
@@ -0,0 +1,52 @@
...
+            "System.IO.Stuff" `matchesGlob` mkGlob "System.IO.*" `shouldBe` True
+        it "should allow ** to match anything on the same level" $
+            "Level1.Level2.Level3" `matchesGlob` mkGlob "Level1.L**2.Level3" `shouldBe` True
+        it "should allow ** to match anything on multiple levels" $
+            "Level1.Level2.Level3.Level4.Level5" `matchesGlob` mkGlob "Level1.L**4.Level5" `shouldBe` True
+
+        describe "Properties" $ do
+            it "should obey: s `matches` glob s with s being any string without '*'" $
+                property $
+                    \(Wildcardless s) -> s `matchesGlob` mkGlob s
```
brandonchinn178, line 46: I believe hspec has a `prop` function built in, so can just do
```suggestion
            prop "should obey: s `matches` glob s with s being any string without '*'" $
                    \(Wildcardless s) -> s `matchesGlob` mkGlob s
```

web/site/pages/config/import-grouping.md
```diff
@@ -0,0 +1,156 @@
...
+  - name: "Specific monads"
+    rules:
+      - glob: Control.Monad.**
+```
+
+Here's an example of using the import grouping feature using the CLI:
+
+```shell
+fourmolu Main.hs --import-grouping by-scope --local-modules SomeInternal.Module1 --local-modules SomeModule
+```
```
brandonchinn178, line 54: Can we move this into the `local-modules` docs?

web/site/pages/config/import-grouping.md
```diff
@@ -0,0 +1,156 @@
...
+- `preserve`: preserve the existing import groups
+- `single`: single group of imports
+- `by-qualified`: one group for unqualified imports, then one group for qualified imports
+- `by-scope`: one group for external imports, then one group for imports targeting modules from the current Cabal project
+- `by-scope-then-qualified`: apply `by-scope` first, then `by-qualified`
+
+Predefined matchers:
+
+- `match: all`: matches all modules
+- `match: local-modules`: matches modules defined in the current Cabal project. Those modules are automatically detected. Developers can add custom modules through the `--local-modules` CLI option.
```
brandonchinn178, line 19: Can we add the following:
```md
See [`local-modules`](/config/local-modules.md) for more information.
```
?

brandonchinn178 (approved):
Thanks for sticking through this work, I really appreciate all your work here. I think we ended up in a really great spot!

src/Ormolu/Config/Types.hs
```diff
@@ -0,0 +1,99 @@
...
+
+instance Aeson.FromJSON ImportGroup where
+  parseJSON = Aeson.withObject "ImportGroup" $ \o ->
+    ImportGroup
+      <$> Aeson.parseFieldMaybe o "name"
+      <*> Aeson.parseField o "rules"
+
+data ImportGroupRule = ImportGroupRule
+  { igrModuleMatcher :: !ImportModuleMatcher,
+    -- | 'Just True' to match qualified declarations, 'Just False' to match unqualified ones and 'Nothing' to match both
```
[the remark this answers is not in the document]
michivi: Right, I missed this, sorry! :sweat_smile: Nice catch!

src/Ormolu/Config/Types.hs
```diff
@@ -0,0 +1,99 @@
...
+        attemptParseModuleMatcher = parseModuleMatcher <|> failUnknownModuleMatcher
+    igrModuleMatcher <- attemptParseModuleMatcher
+
+    qualified <- o .:? "qualified"
+    igrQualifiedMatcher <- case qualified of
+      Just True -> pure MatchQualifiedOnly
+      Just False -> pure MatchUnqualifiedOnly
+      Nothing -> pure MatchBothQualifiedAndUnqualified
+
+    igrPriority <- o .:? "priority" .!= defaultImportRulePriority
```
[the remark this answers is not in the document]
michivi: Done. Also created dedicated definitions so that we won't have hard-coded values everywhere. I've kept `.!=` for now as I thought it was cleaner. Would you rather use `fromMaybe`?

web/site/pages/config/import-grouping.md
michivi: Done

web/site/pages/config/import-grouping.md
michivi: Done

src/Ormolu/Config/Types.hs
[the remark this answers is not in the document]
brandonchinn178: Nope! I forgot that `.!=` is basically an infix version of `fromMaybe`

michivi:
Sorry, thought I had rebased on the upstream `main`, but I was actually stuck on `0.15.0.0`. Had to review everything after conflict resolution, but I think everything is still here.

brandonchinn178:
@michivi Thanks a ton! Lots of great work

PR #406, opened 2024-05-14
Relax upper bound to admit ansi-terminal-1.1.1

mpilgrem:
Motivation: this is blocking the reintroduction of `ansi-terminal-1.1.x` to nightly snapshots. See: https://github.com/commercialhaskell/stackage/pull/7420#issuecomment-2110842180

### DIFF
--- a/fourmolu.cabal
+++ b/fourmolu.cabal
@@ -108,7 +108,7 @@ library
         Cabal-syntax >=3.10 && <3.11,
         Diff >=0.4 && <1,
         MemoTrie >=0.6 && <0.7,
-        ansi-terminal >=0.10 && <1.1,
+        ansi-terminal >=0.10 && <1.2,
         array >=0.5 && <0.6,
         base >=4.14 && <5,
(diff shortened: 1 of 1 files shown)


mpilgrem:
I think the CI is failing for reasons unrelated to the pull request. I suspect it may be because the workflow has not kept up with the change to GitHub-hosted runner `macos-latest` (which does not come with Haskell tools).

andreasabel:
Indeed.  I suggest to add a `haskell-actions/setup` step to install stack.  This action has been surprisingly robust wrt. to the last macOS runner change (praise goes to the original action writers).
- https://github.com/actions/runner-images/issues/9242

juhp:
In the meantime a Hackage revision would also be welcome and helpful, thanks

I can also help with that as a Trustee

brandonchinn178 (approved):
Thanks!

brandonchinn178:
`ansi-terminal` upper bound bumped on Hackage

PR #410, opened 2024-05-24
Preserve function params as single line when constraints or foralls are on another line

### LINKED ISSUE #277: Support single-line function type after multi-line constraints
- [x] Yes, this feature is not relevant to Ormolu

**Describe the current + desired formatting**
```hs
-- input
foo ::
  ( Eq a
  , Show a
  ) =>
  a -> a
```

Desired: input is unchanged

Current:
```hs
foo ::
  ( Eq a
  , Show a
  ) =>
  a ->
  a
```
Same for other `function-arrows` options

lykahb:
Resolves #277 

This adds a new option `newline-inside-type` to control what happens when the function arguments are on one line, and yet the signature has line breaks after constraints or foralls.

This is meant to find the middle ground between a very sparse type where every parameter is on its own line, and a one-liner that is hard to read. This is especially helpful when the constraints and the rest of the signature have comparable sizes.

For example, this enables:
```
mapM ::
  (Traversable t, Monad m) =>
  (a -> m b) -> t a -> m (t b)
```
rather than choosing between
```
mapM ::
   (Traversable t, Monad m) =>
   (a -> m b) ->
   t a ->
   m (t b)

mapM :: (Traversable t, Monad m) => (a -> m b) -> t a -> m (t b)
```

This PR also makes preserving the single lines the default option. While this required changes in the test suite examples (the inputs had many things on a single line), I don't think that this would be a breaking change for the existing codebases that use fourmolu - the existing multi-line signatures wouldn't get affected. I'd be happy to turn it off by default if you prefer.


config/generate.sh
```diff
@@ -5,4 +5,11 @@ set -eux -o pipefail
...
 cd "${HERE}"
 
-runghc -Wall -Werror Generate.hs
+if command -v runghc &> /dev/null; then
+    runghc -- -Wall -Werror Generate.hs
+elif command -v stack &> /dev/null; then
+    stack runghc -- -Wall -Werror Generate.hs
+else
+    echo "runghc and stack could not be found"
+    exit 1
```
lykahb, line 14: A drive-by change. Fourmolu is configured with stack, and yet this script needs a regular GHC installation to run.

brandonchinn178: Thanks! FWIW Fourmolu can be built with both stack and cabal, but this script should work regardless, so thanks for the change!

Nit: It would probably be a tad better as
```bash
RUNGHC=runghc
if command -v stack &> /dev/null; then
  RUNGHC=stack runghc
endif

${RUNGHC} -- -Wall -Werror Generate.hs
```

No need for an explicit error (just error with "runghc not found")

data/examples/declaration/signature/type/context-body-separated-single-line.hs
```diff
@@ -0,0 +1,3 @@
+functionName
```
brandonchinn178, line 1: Don't add tests to `data/examples/` directly; those are all Ormolu files. Either include these in the `data/fourmolu/newline-inside-type/` tests (preferred) or in the dedicated `data/examples/fourmolu/` folder

config/FourmoluConfig/ConfigData.hs
```diff
@@ -228,6 +228,16 @@ allOptions =
         ormolu = HsList [],
         sinceVersion = Just "0.13.0.0",
         cliOverrides = emptyOverrides
+      },
+    Option
+      { name = "newline-inside-type",
```
brandonchinn178, line 233: The name of this option is a little odd to me. It's also not immediately clear how this interacts with the normal Fourmolu behavior. If someone actually wants
```hs
foo ::
  forall x.
  Show x =>
  x ->
  String
```
How do they set this option?

brandonchinn178 (commented):
Initial comments, haven't finished reviewing

config/FourmoluConfig/ConfigData.hs
lykahb: Yeah, it was hard to find a concise name that describes the option behavior, and I'm not quite happy with this name either. My other options were "two-line-signature" (True/False), "constraints-newline-function-args" (with options "preserve-single-line", "multi-line"), "respectful-function-parameters" (True/False, respects that they are single-line, even when other parts of signature aren't).

The "multi-line" option reverts to the normal Fourmolu/Ormolu behavior. So, with an input like this and the `--newline-inside-type=multi-line`, it would break up this input into the output that you posted.
```
foo ::
  forall x.
  Show x =>
  x -> String
```

data/examples/declaration/signature/type/context-body-separated-single-line.hs
lykahb: Moved into the `data/fourmolu/newline-inside-type/`

data/fourmolu/newline-inside-type/input.hs
brandonchinn178, line 1: So I can get a better understanding of this option, can you add the following examples:
1. with `forall`
2. without any forall/constraints
3. with arguments on multiple lines (both for the existing examples and the above examples)

lykahb: Pushed more examples in the test and gave them more meaningful names.

In short, this option makes the formatter respect when the tail of the signature is one line, even if there are multi-line constraints or forall's preceding it.

brandonchinn178 (commented):
Thanks for adding the extra examples, it was really helpful. So it seems like it's just adding another "single line" case; previously, Fourmolu kept everything on a single line if it already is and multiline if not, and this option adds another possibility of "put foralls, constraints, and type on new lines, but each individually on a single line".

I would instead rename the option and describe it as follows:
* `single-line-types`
    * `auto` - default; single line when everything is already on one line, allows foralls/constraints/type to be treated separately
    * `only-if-already-single` - single line only if everything is already on one line, including foralls/constraints. old behavior, mostly to get parity with ormolu
    * (optional) `always` - always do single line
    * (optional) `never` - always do multi line

What do you think?

lykahb:
@brandonchinn178 Thanks for the suggestion, I renamed the option and the values. I left out the "always" and "never" options, though. I couldn't get the "never" option work, and it's not clear how "always" would work with the haddock comments on the types.

brandonchinn178 (approved):
This shaped up to be a really cool feature! Thanks @lykahb!

You need to fix one linting failure before merging.

levinean (approved):

web/site/pages/config/single-line-types.md
```diff
@@ -0,0 +1,24 @@
...
+{ "single-line-types": "auto" }
+```
+```fourmolu-example-tab
+only-if-already-single
+{ "single-line-types": "only-if-already-single" }
+```
+
+## See also
+
+* [`respectful`](/config/respectful)
```
brandonchinn178, line 24: Sorry, one super minor question: how does this option relate to `respectful`?

lykahb: The `respectful` option avoids changing line breaks from how the user inputted them. I think that the `single-line-types` option is similar in a sense that fourmolu respects the absence of line breaks in a type.

brandonchinn178: Makes sense. thanks!

web/site/pages/config/single-line-types.md
```diff
@@ -0,0 +1,24 @@
...
+
+$info$
+
+The `auto` option keeps parameters in a single line, ignoring whether foralls/constraints are on a different line. This allows type to be treated separately from foralls/constraints, and use newlines before type without forcing the whole signature to be formatted multi-line.
+
+## Examples
+
+```fourmolu-example-input
+(<$>) :: Functor f =>
+  (a -> b) -> f a -> f b
```
brandonchinn178, line 11: This is failing the build, probably due to the `$` not being escaped. Either escape it, or use a different function as an example

michaelpj:
This looks nice! I just had some bikeshedding thoughts about the name. The way this makes sense to me is that currently we treat a type signature as a single entity to be printed single-line or multi-line. With this setting we instead consider the context and the rest of the types as sub-parts of the signature, which can have an independent decision about whether to print single- or multi-line.

So a possible different option name could be something like `independent-signature-parts: allows the context and type inside a type signature to make independent choices for single- or multi-line printing`. Maybe that's not clearer, though :laughing:

brandonchinn178:
@michaelpj Thanks for the suggestion! One issue I have with that option name is that the signature parts are already kinda independent. The following signatures are both fixed points for Fourmolu:
```hs
f ::
    (Monad m, Show a) =>
    a ->
    m String

g ::
    ( Monad m
    , Show a
    ) =>
    a ->
    m String
```
So the constraint part of the signature can be single/multi-line independently of the args part of the signature. The problem here is specifically the args part not having the same behavior.

But this discussion raised another question for me: why would anyone not want this behavior? Looking back at #277, I said this issue was not relevant for Ormolu, but I'm not sure why I said that. Ormolu also has the same behavior. So here's what I'll do:
1. I'll submit a PR for this in Ormolu, see if they want it
2. If not, let's name this option `allow-singleline-args-in-multiline-type` that, similar to `respectful`, defaults to `true` but can be turned off only for Ormolu compat. But we don't expect anyone to disable this.
    * If we do this option, I'll merge the PR as-is and make the name change myself, to give @lykahb a bit of a break 😄

michaelpj:
Sounds plausible to me!

brandonchinn178:
@lykahb Thanks a ton for your work here! The change was merged into Ormolu here: https://github.com/tweag/ormolu/pull/1125, so we'll get this change after the next Ormolu release. Are we good to close this PR?

lykahb:
Thanks for porting this change to Ormolu. I expected them to be more conservative about changing the formatting.

PR #402, opened 2024-02-28
Fix indentation in multiline tuple patterns

brandonchinn178:
This formatting behavior was discovered in https://github.com/fourmolu/fourmolu/issues/400. It's partially an issue in Ormolu (https://github.com/tweag/ormolu/pull/1101), but it's more egregious with leading commas.

Hold off on merging this until https://github.com/tweag/ormolu/pull/1101 is resolved; if it's accepted upstream, we can hold off on this and rebase after merging Ormolu in

PR #418, opened 2024-06-02
Reduce indentation of multiline brackets

brandonchinn178:
An alternate fix for #402 that only indents multiline bracketed expressions by a single space, which is what's minimally necessary to avoid parser errors.

~~Upstream fix: https://github.com/tweag/ormolu/pull/1116~~ Rejected upstream, we'll just make the change on our end

### DIFF
--- a/changelog.d/multiline-brackets.md
+++ b/changelog.d/multiline-brackets.md
@@ -1,0 +1,1 @@
+* Reduce indentation of multiline bracketed expressions
--- a/compat-tests/cabal.diff
+++ b/compat-tests/cabal.diff
@@ -77,6 +77,59 @@ index b03b1b9..a876a5c 100644
 +            prettyFieldGrammar cabalSpecLatest buildInfoFieldGrammar bi
           | (name, bi) <- ex_bis
           ]
+diff --git a/Cabal-syntax/src/Distribution/Types/PackageDescription.hs b/Cabal-syntax/src/Distribution/Types/PackageDescription.hs
+index f8f8431..7311ebd 100644
+--- a/Cabal-syntax/src/Distribution/Types/PackageDescription.hs
++++ b/Cabal-syntax/src/Distribution/Types/PackageDescription.hs
+@@ -488,7 +488,7 @@ instance L.HasBuildInfos PackageDescription where
+         a22
+         a23
+         a24
+-      ) =
++     ) =
+       PackageDescription a1 a2 a3 a4 a5 a6 a7 a8 a9 a10 a11 a12 a13 a14 a15 a16 a17 a18 a19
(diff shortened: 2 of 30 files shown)


brandonchinn178:
@georgefst Thoughts on this? It's a breaking change, but I can't think of a reason anyone would want the old format

georgefst:
I'm happy with occasional breaking changes which are an obvious improvement, but unfortunately some outputs like [this](https://github.com/fourmolu/fourmolu/pull/418/files#diff-0022900b8678d90bfd5bb93464be301c8bb0469376d11b3f2c38d8d093c1fb83) look worse than before. Would it be easy to align `|` with `[` there?

brandonchinn178:
@georgefst to be clear, you're referring to `data/fourmolu/let-style/output-LetAuto-InLeftAlign-indent=4.hs`? There are two list comprehensions in that diff. The second one has to be shifted, or else it's a parse error. The first one is related, but slightly misleading; see below.

There are two issues going on here:
1. Multiline bracketed expressions in certain contexts are indented a full indentation. This is fine in Ormolu because it's just 2 spaces, but it looks weird with more indentations. In these contexts, the only purpose to the indentation is to avoid putting subsequent lines in the same column as the first line, which can cause parse errors. So in this PR, I'm changing this from a full indentation to a single space. Ormolu may or may not upstream this, but I think this is fine as a Fourmolu-only change.
2. List comprehensions are currently formatted the same as normal bracketed expressions, but this is an artifact from how Ormolu styles things. I have a fix here https://github.com/tweag/ormolu/pull/1119, which separates list comprehensions styling from other bracketed expressions.

So ignoring list comprehensions, do the other outputs look like an improved format to you?

georgefst:
> to be clear, you're referring to data/fourmolu/let-style/output-LetAuto-InLeftAlign-indent=4.hs?

Ah, probably, I can't remember now! Not sure what happened with that link.

Anyway, most of these changes look positive, but I'm afraid I can't sign off on stuff like this from `data/examples/declaration/value/function/pattern/multiline-case-pattern-four-out.hs`:
```diff
f xs = case xs of
    [ a
-       , b
-       ] -> a + b
+    , b
+    ] -> a + b
```

This is a regression in terms of "number of lines whose number of leading spaces is a multiple of the configured indentation step", which I'm sure most people would agree is an important metric. Of course, losing another space results in a parse error. So really the _ideal_ fix here might involve patching GHC...

brandonchinn178:
Ah sorry forgot to update. This should be unnecessary with https://github.com/fourmolu/fourmolu/pull/429

georgefst:
Excellent! I hadn't caught up as far as #429 yet.

PR #420, opened 2024-06-02
Update help text column widths

### LINKED ISSUE #31: Help text is ugly
Perhaps my fault for choosing some rather long field names... But really this is an `optparse-applicative` issue: https://github.com/pcapriotti/optparse-applicative/issues/406

```
  --start-line START       Start line of the region to format (starts from 1)
  --end-line END           End line of the region to format (inclusive)
  --indentation WIDTH      Number of spaces per indentation step (default 4)
  --comma-style STYLE      How to place commas in multi-line lists, records etc: 'leading' (default) or 'trailing'
  --indent-wheres BOOL     Whether to indent 'where' bindings past the preceding body (rather than half-indenting the
                           'where' keyword) (default 'false')
  --record-brace-space BOOL
                           Whether to leave a space before an opening record brace (default 'false')
  --diff-friendly-import-export BOOL
                           Whether to make use of extra commas in import/export lists (as opposed to Ormolu's style)
                           (default 'true')
  --preserve-spacing BOOL  Give the programmer more choice on where to insert blank lines (default 'true')
  --haddock-style STYLE    How to print Haddock comments (default 'multi-line')
  FILE                     Haskell source files to format or stdin (default)
```

brandonchinn178:
Resolves #31

Here's the before:
Screenshot 2024-06-02 at 1 59 45 PM

Here's the after (terminal size = 100):
Screenshot 2024-06-02 at 1 59 56 PM

### DIFF
--- a/DEVELOPER.md
+++ b/DEVELOPER.md
@@ -112,6 +112,8 @@ To release a new version, do the following workflow:
 
     1. Audit `web/site/` docs
 
+    1. Check that `--help` looks good
+
 1. Create PR as usual and merge into `main`
     1. In the `check_sdist` CI job, check the output of the `stack sdist` step for any warnings.
 
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -31,6 +31,7 @@ import Ormolu.Terminal
 import Ormolu.Utils (showOutputable)
 import Ormolu.Utils.Fixity
 import Paths_fourmolu (version)
+import System.Console.Terminal.Size qualified as Terminal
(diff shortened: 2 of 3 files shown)


brandonchinn178:
@georgefst does this look better for you?

georgefst:
> @georgefst does this look better for you?

I'm a bit late but, yes, absolutely!

(It would be nicer if it were automatic, rather than having to hardcode 35, but this would require upstream support and they've argued it would be difficult.)

Issue #428, opened 2024-08-08
Multiline expr starts one column less than Ormolu


tomjaguarpaw:
**Is your bug specific to Fourmolu?**
- [x] Yes, I've verified that Ormolu does not have this bug

**Did you try it on the web app?**
- [x] Yes, I've verified that https://fourmolu.github.io can reproduce this bug

**Did you search for existing issues?**
- [x] Yes, ~there are no open or closed issues related to my issue~ "ormolu less indentation" throws up nothing obvious

**Describe the bug**

Ormolu formats like this

```.hs
f :: T
f =
  something
    <> ( myFun $
           Just t
       )
```

whereas fourmolu formats like this, with indentation 2:

```.hs
f :: T
f =
  something
    <> ( myFun $
          Just t
       )
```

i.e. `Just` is indented one level less. I've suggested to my team to switch to fourmolu because we don't like the latest Ormolu choice of wrapping singleton constraints in parens, but it seems the amount of churn we'll get by switching to fourmolu is even bigger because of this difference. Is this difference really intended? If so, why?

brandonchinn178:
Yeah, that's kind of expected 🫤 Also see
https://github.com/fourmolu/fourmolu/commit/e9e9ad63dd938868a6fe968c70c592e5cff346b0

The problem is that Fourmolu ensures that a line always begins on a multiple of the indentation, rounding down, which IMO is the better behavior. I thought I had ticketed this in Ormolu, but I can't find it. Let me see what Ormolu says

tomjaguarpaw:
Hmm, OK, I guess this rules out fourmolu for us then :(

brandonchinn178:
More context: https://github.com/fourmolu/fourmolu/pull/37#issuecomment-701733625

This comment is a bit inaccurate, though; it's not at that column because that's where it would've been on the previous line, but because it's a 2-space indentation starting from the expression inside the parentheses:
```hs
f =
  False
    || ( something
           >= other
       )
```

I think really, we want this rule to be "if indentation > 2, indent from parentheses, otherwise indent from the inner expression". Let me take a look...

brandonchinn178:
hm doesn't seem trivial, and for something that's arguably not a big deal... will leave the issue open and punt for now

PR #429, opened 2024-08-09
Only apply indentation hack for indentation > 2

brandonchinn178:
We have logic in `inci` to round down to the closest multiple of the indentation before increasing the indentation, in the case of parenthesized expressions, which shift the indentation off the indentation column.

```hs
-- Ormolu behavior:
( Maybe
    Int -> -- <- 'Int' is one indentation (2 spaces) from the column 'Maybe' starts on
  String
)

-- If Fourmolu just naively changed indentation to 4:
( Maybe
      Int -> -- <- 'Int' is one indentation (4 spaces) from the column 'Maybe' starts on,
  String     -- but it now starts on column 6, because of the "( ", so all subsequent
)            -- newlines will be off the indentation column

-- Fourmolu's behavior today, rounding down (indentation=4):
( Maybe
    Int -> -- <- The start of 'Maybe' is on column 2, so we round down to a multiple
  String   -- of 4 (0), then increase the indentation to 4
)
```

This hack works when the parentheses starts on the indentation column, but can cause odd formatting when the parentheses starts off the indentation column (https://github.com/fourmolu/fourmolu/issues/428)

```hs
-- Ormolu behavior:
something
  <> ( foo
         bar -- <- 'bar' is one indentation (2 spaces) from the start of 'foo', but 'bar'
         baz -- and 'baz' are now on an odd column, due to being shifted 5 characters
     )       -- due to "<> ( "

-- Fourmolu behavior today, rounding down (indentation=2):
something
  <> ( foo
        bar -- <- 'bar' is now only 1 space from 'foo' to keep it on the indentation column,
        baz -- which is different from what Ormolu does at 2 space indentation
     )
```

So we'll hack the hack by only applying the hack if indentation > 2. A better solution might involve something like tracking the current indentation separately from the next indentation level, but I don't want to think about that right now

### DIFF
--- a/changelog.d/indent.md
+++ b/changelog.d/indent.md
@@ -1,0 +1,1 @@
+* Fix indentation for parenthesized expressions that start off the indentation column ([#428](https://github.com/fourmolu/fourmolu/issues/428))
--- a/compat-tests/cabal.diff
+++ b/compat-tests/cabal.diff
@@ -24,8 +24,42 @@ index 47d4667..ffd1713 100644
    , module Distribution.Types.GenericPackageDescription
  
      -- * Components
+diff --git a/Cabal-syntax/src/Distribution/PackageDescription/Configuration.hs b/Cabal-syntax/src/Distribution/PackageDescription/Configuration.hs
+index 9a9ba2d..f88eb22 100644
+--- a/Cabal-syntax/src/Distribution/PackageDescription/Configuration.hs
++++ b/Cabal-syntax/src/Distribution/PackageDescription/Configuration.hs
+@@ -117,12 +117,12 @@ parseCondition = condOr
+     cond =
+       sp
+         >> ( boolLiteral
+-              <|> inparens condOr
+-              <|> notCond
+-              <|> osCond
+-              <|> archCond
+-              <|> flagCond
+-              <|> implCond
++               <|> inparens condOr
++               <|> notCond
++               <|> osCond
++               <|> archCond
++               <|> flagCond
++               <|> implCond
+            )
+     inparens = between (P.char '(' >> sp) (sp >> P.char ')' >> sp)
+     notCond = P.char '!' >> sp >> cond >>= return . CNot
+@@ -465,8 +465,8 @@ finalizePD
+   -- ^ Additional constraints
+   -> GenericPackageDescription
+   -> Either
+-      [Dependency]
+-      (PackageDescription, FlagAssignment)
++       [Dependency]
++       (PackageDescription, FlagAssignment)
+   -- ^ Either missing dependencies or the resolved package
+   -- description along with the flag assignments chosen.
+ finalizePD
 diff --git a/Cabal-syntax/src/Distribution/PackageDescription/PrettyPrint.hs b/Cabal-syntax/src/Distribution/PackageDescription/PrettyPrint.hs
-index b03b1b9..a876a5c 100644
+index b03b1b9..4039245 100644
 --- a/Cabal-syntax/src/Distribution/PackageDescription/PrettyPrint.hs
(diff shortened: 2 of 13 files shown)


brandonchinn178:
@Kleidukos This changes quite a few lines in the Cabal codebase, but IMO it's an improvement to the formatting. What do you think? Would you prefer this be gated behind an option?

Kleidukos:
@geekosaur can I let you check this out? :heart_hands:

ulysses4ever:
It'd be easier to opine if the three paragraphs of the PR description were accompanied by an example. 

Cabal still uses fourmolu 0.12 so I don't see how it could affect it. One day it should move on though. It'd be great if changes in new versions of fourmolu were guarded by flags always (apologies if this is already the case but I think it wasn't always the case in the past). Opt in or opt out is a judgement call of course. But there should be a way to move to a newer version without reformating the whole code base. And that way should be explained in release notes...

geekosaur:
Looks fine to me. (Note: as I read it, 1-space indents have been expanded to 2-space, because the leading "+" or "-" is " " for unchanged lines. This makes diffs readable _unless_ they're of tab-indented lines.)

brandonchinn178:
@ulysses4ever Thanks for the comment, I updated the PR description with examples.

On one hand, I agree, people wanting to upgrade GHC will need to upgrade fourmolu, and it'd be annoying to have to include reformatting the codebase in order to upgrade GHC version. On the other, maintaining flags to maintain backwards compatibility in perpetuity doesn't sound like a fun time (this particular instance is fairly small, but speaking generally).

What do you think about some policy like:
* All backwards incompatible changes should be opt-out
* Flag is guaranteed to exist for at least one version of Fourmolu that supports GHC X.Y+1
* When flag is used, Fourmolu shows warning to migrate
* Fourmolu is free to remove flag if flag was introduced in the previous GHC version

ulysses4ever:
Thank you Brandon for engaging. Couple points.

1. I would welcome any policy if there's none currently. I think anything formal will be an improvement.
2. I don't use fourmolu enough to have a good intuition about precise terms.
3. The proposal you put seems to lean conservative esp. concerning updates for newer GHCs and dropping those flags,  but this is fine because it's your call and also see (1) above...

The issue with cabal is that big swings in output of newer fourmolu's invalidate a lot of PRs, and we always have a good bunch open ones... Given that cabal has to follow GHC pretty closely, the policy you propose doesn't give us a ton of time to adjust (if we want to update fourmolu regularly). It's half a year that takes GHC X.Y+1 to release normally (I hope I got it right?) worst case (and I know that GHC releases get delayed all the time but if we need a number...).

brandonchinn178:
That's a good point. What frequency would Cabal be open to reformatting the codebase? Once a year? I'd be happy to make it "remove a flag after GHC X.Y+1 or 1 year, whichever comes later"

brandonchinn178:
I'll already be at that meeting, for a different reason 😁

How often does Cabal need to update what version of GHC it builds with? You can't stay pinned on 0.12 if you need to upgrade the version of GHC that builds fourmolu, or if cabal uses newer syntax.

If that's not an issue, then I think we can go back to just the GHC X.Y+1 policy.

Mikolaj:
I'm inclined to haggle to make it less common than once a year. :) But let's find a compromise.

Kleidukos:
@brandonchinn178 see you in 5 minutes then :)

brandonchinn178:
Per the newly merged breaking changes policy, I'm tempted to merge this in. The current implementation violates the policy:

> It will always be possible to replicate Ormolu's formatting with a suitable `fourmolu.yaml`

So this one-time breaking change would make us consistent with Ormolu again

georgefst:
Very nice! We should have done this years ago.

I was never very happy with #37. No coincidence that it stems from the period when there was no one around to review my code...

Issue #363, opened 2023-08-29
Guards indentation


epoberezkin:
**Describe the option you'd like to add**
* e.g. name of option, valid values, etc.
* What option setting would keep current behavior?

Proposed option: `indent-guards` with values `true` (current behaviour) and `false` (ormolu)

**Show a code example exemplifying the different option settings**

fourmolu (with `indentation: 2`):

```
f x y z
  | x =
      y
  | otherwise =
      z
```

ormolu (and fourmolu with `indent-guards: false`):

```Haskell
f x y z
  | x =
    y
  | otherwise =
    z
```

epoberezkin:
In general, given that it is a fork, why the starting point wasn't taken to keep ormolu behaviour as default and add options (and, possibly, option sets) to change it?

brandonchinn178:
Yes, I contributed that change to ormolu 🙂 It wasn't "for no apparent reason", I explain the reasoning in the issue.

Do you actually prefer it unindented, or do you just want to avoid reformatting your code? If the latter, you could just keep your fourmolu/ormolu version pinned, right? Or if you're concerned about git blame, you can add it to [`.git-blame-ignore-revs`](https://github.blog/changelog/2022-03-24-ignore-commits-in-the-blame-view-beta/)?

Regardless, feel free to submit a PR.

brandonchinn178:
Personally, I would not want to set a precedent of adding a Fourmolu option to revert any change Ormolu makes. Fourmolu is committed to being very conservative about formatting changes (just like Ormolu, as Mark mentioned in the ormolu issue), so breaking changes should occur less and less frequently over time.

Plus, this is a one-time change. You can just reformat your codebase, optionally add the commit to `.git-blame-ignore-revs`, and be done in (hopefully) less than a minute. For people working in your codebase, the new format is (hopefully) easier to read than the old format, so reverting the change adds a nonzero cost every time someone reads an expression after a guard.

As a maintainer, we would have to pay the cost of maintaining this toggle forever. If Ormolu ever makes changes around this area, merging Ormolu into Fourmolu will cause a conflict we'll have to resolve. This seems hard to justify if no one actually likes that style, but you just want it for backwards compatibility of this one version bump.

Thinking about this more, I don't think it's worthwhile to have this option in Fourmolu. You're always free to fork Fourmolu (or Ormolu) and revert the change in your own repo. I hope that makes sense; please let me know if you feel I'm being unreasonable

Issue #345, opened 2023-07-08
Stable formatting behaviour across `fourmolu` releases


ffaf1:
Hello `fourmolu` devs,

I am a [`cabal`](https://github.com/haskell/cabal) contributor. `cabal` recently [picked `fourmolu` as its automatic formatter](https://github.com/search?q=repo%3Ahaskell%2Fcabal+fourmolu&type=commits).

I write here to you see how to iron out some problems we encountered in using fourmolu. Story time:
- on `cabal` CI `fourmolu` is pinned to `0.12`;
- a developer, while working on a patch, installed `fourmolu` 0.13 on their machine;
- this caused some mismatch and CI refusing the patch after pushing.

Of course the answer to this problem is to specify which fourmolu version to use ([patch](https://github.com/haskell/cabal/pull/9094)) but I am a bit worried. E.g.:
- Some `GHC`/`base` eventually will not allow `fourmolu-0.12` to be built, adding friction to contributors.
- Maybe a bug was fixed in `0.13` but I cannot switch to it becasue it adds additional formatting rules.
- Eventually the project will have to switch to a newer fourmolu, having yet another whole-codebase big patch to apply and possibly creating havoc for unmerged PRs.

Hence me writing here. What I would love to have is a stable formatting behaviour accross `fourmolu` versions, but I am quite open to listen to alternatives: `fourmolu` is used in many codebases and for sure they encountered similar problems and found a sensible way forward which works for them.

Thanks again for this very useful formatter!

brandonchinn178:
Thanks for writing this issue! A few comments:

* What exactly was the change? AFAIK 0.13.0.0 shouldn't include any major changes.
* Fourmolu does aim to be stable now. All formatting rules added recently, and all formatting rules added in the foreseeable future should be backwards compatible
* That being said, Fourmolu does fork Ormolu, so if Ormolu makes a breaking change, there's not much we can do there.

As an example, when I upgraded from Fourmolu 0.7 to 0.12 in my personal projects with the same config, the only changes were adding parentheses around single constraints, a change Ormolu made. So even that change could be avoided by using `single-constraint-parens: never`. From the README:

> Fourmolu is in a fairly early stage of development. The implementation should be as stable as Ormolu, as it only makes minimal changes, and is extensively tested. But the default configuration style may change in some minor ways in the near future, as we make more options available. It will always be possible to replicate the old default behaviour with a suitable fourmolu.yaml.

[...]

PR #430, opened 2024-08-20
RFC: Add policy for breaking changes

brandonchinn178:
## Context + Motivation

While Fourmolu remains configurable, certain changes may still be breaking, e.g. if Ormolu makes a breaking change, or if no one would use the old format if not for old Fourmolu behavior. Generally, people should always be pinning a Fourmolu version in a project, but sometimes upgrades are required, e.g. when upgrading a project to use a newer GHC version. Simply upgrading a codebase to use a newer GHC can be a difficult task (e.g. https://github.com/haskell/core-libraries-committee/issues/12#issuecomment-971794515), so it can be frustrating when you realize you also have to reformat the codebase because upgrading to a version of Fourmolu that supports the new GHC version includes breaking changes. Somewhat relevant discussions: https://github.com/fourmolu/fourmolu/issues/363, https://github.com/fourmolu/fourmolu/issues/345

At the same time, we don't want to support backwards compatibility flags indefinitely, as it makes merging Ormolu changes increasingly difficult. But perhaps there's a middle ground. Regardless, we should make such a policy explicit.

## Options

There are two possible policies for breaking changes I can see Fourmolu adopting:

1. Status quo: Fourmolu will do its best to avoid breaking changes, but it may still happen. When upgrading Fourmolu, you might have to reformat at the same time. If you're upgrading GHC, you should probably upgrade Fourmolu first + reformat, then finish upgrading GHC.
2. This PR: Fourmolu will always guard breaking changes (even ones from Ormolu) until Fourmolu releases a version on the next GHC version.

The benefit of policy (2) is it gives users a transition period of at least 6 months (given GHC's release timeline is every 6 months). The downside is it adds the maintenance burden of a new flag for every breaking Ormolu change in the next 6 months. The question here is: _**Is it worth having this transition period if it's going to be required to reformat at some point anyway?**_

## User feedback - Cabal

[Cabal](https://github.com/haskell/cabal) is probably the biggest project using Fourmolu, so we added a backwards compatibility test in #346 to surface breaking changes against the Cabal codebase. In a recent PR (#429), we saw a large reformat in the Cabal codebase, which led to a discussion in the comments with Cabal maintainers.

To align on the issue, I (Brandon) joined the biweekly Cabal maintainers' meeting on 2024-08-15, and explained the issue. Namely, that certain reformattings don't make sense to allow configuring indefinitely. In this particular case, the current formatting is inconsistent with Ormolu, and this brings Fourmolu back to consistency.

In the meeting, the group agreed that these sort of breaking changes make sense, and are reasonable. They resolved to upgrade Fourmolu on a regular basis (every 6 months?), and to accept the reformat regularly, even adding to their release checklist to consider upgrading Fourmolu + reformat before making the branch cut for a release. So regardless whatever policy Fourmolu adopts, it probably won't affect Cabal.

## PR TODO's

If policy (2) is accepted, finish doing the following:

- [ ] Implement `backwards-compat-single-line-args`
- [ ] Implement `backwards-compat-list-comp-indent`
- [ ] Update diffs in `compat-tests/` to turn on backwards compat options

If policy (1) is accepted, close this PR, but open a new one that explicitly documents the policy.

### DIFF
--- a/.github/workflows/pr-review-comment.yml
+++ b/.github/workflows/pr-review-comment.yml
@@ -23,4 +23,5 @@ jobs:
             - [ ] Diff files in `compat-tests/` either:
                 1. Have NOT been changed, or
                 2. Modifies `fourmolu.yaml` with an appropriate option to keep the same formatting, or
-                3. Modifies Haskell files with new formatting (ONLY as a last resort)
+                3. Modifies Haskell files with new formatting
+                    * ONLY as a last resort - see [Breaking changes policy](https://github.com/fourmolu/fourmolu#breaking-changes-policy)
--- a/README.md
+++ b/README.md
@@ -18,6 +18,7 @@
 * [Troubleshooting](#troubleshooting)
     * [Operators are being formatted weirdly!](#operators-are-being-formatted-weirdly)
 * [Limitations](#limitations)
+* [Breaking changes policy](#breaking-changes-policy)
 * [Contributing](#contributing)
 * [License](#license)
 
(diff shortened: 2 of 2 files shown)


Bodigrim:
I don't see why Cabal could not stick to a given Fourmolu release for a very long time, 3-5 years. Their codebase is kept compatible with fairly old GHC releases, so there is no new syntax to parse which could force upgrading Fourmolu.

I think the proposed policy is a laudable commitment, but I'm cautious that it can put too much strain on maintainers of Fourmolu. In my view status quo is good enough.

brandonchinn178:
cc @eopberezkin Would be great to get your thoughts

brandonchinn178:
cc @epoberezkin Would be great to get your thoughts

michaelpj:
As far as I'm concerned there would be an issue in one of the following cases:
1. the change requires significant manual action from me
2. multiple people in my team were using different versions and so got different output
3. the change was forced on me at the same time as something else, making that more difficult

1 is I think pretty unlikely, 2 I am happy accepting as "my problem" as a user. 3 seems like it could be annoying if e.g. a GHC upgrade also forced a fourmolu upgrade. So I think the most useful thing would be to avoid that, i.e. try and make sure that the fourmolu release that adds support for GHC X is *not* a breaking change in any other way. 

Adding transition periods doesn't really help much if you are pinning your version of fourmolu. So long as you can manage the transition by stepping between versions, it doesn't really matter if there

Bodigrim:
> 3 seems like it could be annoying if e.g. a GHC upgrade also forced a fourmolu upgrade.

I think this is extremely rare: GHC upgrade could force an upgrade of `fourmolu` only if you immediately start using a new syntax extension. Otherwise you can upgrade to a new GHC, but keep an old, pre-built binary of `fourmolu`.

brandonchinn178:
@Bodigrim That's a good point.

Given that we have alignment with Cabal, and comments here and on [Discourse](https://discourse.haskell.org/t/rfc-fourmolu-breaking-changes-policy/10197) don't bring up anything concerning, we'll stick with the status quo. I'll update this PR with the new breaking changes policy

Issue #298, opened 2023-02-23
space after parenthesis


dark-ether:
**Space after Parenthesis**
currently code like 
```haskell
bracket 
  (allocate some stuff)
  cleanup
  (\s  -> do
    action1
    action2
    action3 
  )  
```
will be formatted as 
```haskell
bracket 
  ( allocate some stuff)
  cleanup
  ( \s  -> do
      action1
      action2
      action3 
  )  
```
the problem is that the extra space changes how the lines in the do block are indented adding an extra space
i would like an option to turn off the extra space after the parenthesis
* What option setting would keep current behavior?
true

brandonchinn178:
Seems reasonable to me, since it does add an additional indentation to parenthesized blocks. PR welcome!

georgefst:
This is a bit confusing to read, especially since the outputs seem to contain extra spaces that Fourmolu would remove. And it wasn't immediately clear that this is only an issue when using 2-space indentation.

Am I correct in thinking that you want:

```hs
def =
  bracket
    (allocate some stuff)
    cleanup
    (\s -> do
      action1
      action2
      action3
    )
```

instead of the current:

```hs
def =
  bracket
    (allocate some stuff)
    cleanup
    ( \s -> do
        action1
        action2
        action3
    )
```

? Where this isn't specific to lambdas, but applies to any expression within multi-line parentheses?

If so, seems reasonable. I'm sure we'd accept a PR.

brandonchinn178:
I think related to https://github.com/tweag/ormolu/issues/891?

PR #431, opened 2024-08-27
Add close-bracket-style option

eownerdead:
Maybe related to #298.

### DIFF
--- a/config/FourmoluConfig/ConfigData.hs
+++ b/config/FourmoluConfig/ConfigData.hs
@@ -90,6 +90,16 @@ allOptions =
         cliOverrides = emptyOverrides
       },
     Option
+      { name = "close-bracket-style",
+        fieldName = Just "poCloseBracketStyle",
+        description = "How to place close brackets in multi-line lists, records, etc.",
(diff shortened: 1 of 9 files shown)


brandonchinn178:
Thanks for submitting a PR! This is an interesting suggestion; I've never seen Haskell code like this in the wild. I'm leaning towards letting this sit until more people have asked for this style.

If you'd like this style, please add a 👍 vote to the PR. If this gets a good number of votes, we can look at merging it.

georgefst:
It's not totally unique, unless [this](https://github.com/cdepillabout/pretty-simple/issues/43#issuecomment-489263342) is @eownerdead's alter ego. Anyway I ended up [implementing it](https://github.com/cdepillabout/pretty-simple/pull/72/files), although I suppose the use case there is for huge expressions that one wouldn't often define as literals.

I don't see any particular reason not to merge this, given how trivial the implementation is. Although I would like to see a test covering nested lists, or a list within a record, to show the brackets bunching up correctly.

> Maybe related to https://github.com/fourmolu/fourmolu/issues/298.

I don't think so.

PR #433, opened 2024-10-06
Sort constraints in contexts

michaelpj:
This adds some simple sorting of constraints. There are a few potential deficiencies:

1. It doesn't sort nested constraint tuples. We could solve this by carrying around some information about whether we are in a constraint context in `p_hsType`, but that seeemed moderately disruptive.
2. It doesn't handle constraint tuple synonyms. I think this is unavoidable since we can't know they're constraint tuples without type information.
3. The sorting is very crude, just sorting on the `Outputable` representation of  the type. I think this will more-or-less doe what people expect, but perhaps it's too expensive?

Fixes #247

### DIFF
--- a/data/fourmolu/sort-constraints/input.hs
+++ b/data/fourmolu/sort-constraints/input.hs
@@ -1,0 +1,43 @@
+module Main where
+
+f1 :: (Show a, Eq a) => a
+f1 a = a
+
+f2 ::
+  (Show a
+  , Eq a)
+  => a
+f2 a = a
+
+-- We don't currently realise that the inner tuple type
+-- is a constraint tuple.
+f3 ::
+  ( Generic a
+  , (Show a, Num a)
+  , Eq a)
+  => a
+f3 a = a
+
+f4 ::
+  (Show a, Num a)
+  =>
+  Eq a
+  => a
+f4 a = a
+
+f5 ::
+  (Show a
+  -- an interloping comment
+  , Eq a)
+  => a
+f5 a = a
+
+class (Show a, Eq a) => Class1 a where
+
+data (Generic a, (Show a, Eq a)) => A a
+
+deriving instance (Show a, Eq a) => Class1 Int
+
+-- We can't know this is a constraint tuple type rather than a normal
+-- tuple type without type information
+type MyConstraints a = (Show a, Eq a)
--- a/src/Ormolu/Printer/Meat/Type.hs
+++ b/src/Ormolu/Printer/Meat/Type.hs
@@ -26,6 +26,7 @@ where
 import Control.Monad
 import Data.Choice (pattern With, pattern Without)
 import Data.Functor ((<&>))
+import Data.List (sortOn)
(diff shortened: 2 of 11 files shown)


data/fourmolu/sort-constraints/input.hs
```diff
@@ -0,0 +1,40 @@
...
+  (Show a
+  , Eq a)
+  => a
+f2 a = a
+
+-- We don't currently realise that the inner tuple type
+-- is a constraint tuple.
+f3 ::
+  ((Show a, Num a)
+  , Eq a)
```
brandonchinn178, line 16: What happens if its `(Eq, (Show, Num))`? Does it still sort the Show above? Can we add a test case for this?

src/Ormolu/Printer/Meat/Type.hs
```diff
@@ -265,10 +266,14 @@ hasDocStrings = \case
   _ -> False
 
 p_hsContext :: HsContext GhcPs -> R ()
-p_hsContext = \case
-  [] -> txt "()"
-  [x] -> located x p_hsType
-  xs -> parens N $ sep commaDel (sitcc . located' p_hsType) xs
+p_hsContext ctx = do
+  shouldSort <- getPrinterOpt poSortConstraints
```
brandonchinn178, line 270: Why not keep the lambda case and just read the opt in the one branch?

web/site/pages/config/sort-constraints.md
```diff
@@ -0,0 +1,26 @@
+# `sort-constraints`
+
+$info$
+
+This option determines wehether or not to sort constraints.
```
brandonchinn178, line 5: `s/wehether/whether/`

brandonchinn178 (commented):
showOutputable is probably fine as a first pass

data/fourmolu/sort-constraints/input.hs
michaelpj: Hmm interesting, it works but I get an "AST differs before and after formatting" error. I probably need to do something similar to what I did before and make the AST comparison not care about ordering here.

michaelpj: I am confused about why I didn't get that in other places though!

michaelpj: Fixed.

brandonchinn178: Is there a test case for this? I only see ((Show, Num), Eq), not (Eq, (Show, Num))

brandonchinn178 (approved):

data/fourmolu/sort-constraints/input.hs
michaelpj: https://github.com/fourmolu/fourmolu/pull/433/files#diff-3f701a29b289a3e81adc2f3677773dd06a2607f45ef1e21c5a95bac9ef0aa3b5R14

has it in the middle

PR #436, opened 2024-10-10
Remove unused `getDotOrmoluForSourceFile` and simplify call sites

georgefst:
I noticed while poking around that this function is actually dead, because we don't use `.ormolu` files, since that config goes in `fourmolu.yaml` instead. So, instead of just effectively hardcoding `--no-dot-ormolu`, I've removed the field internally, which opens up a cascade of simplifications. I haven't thought too heavily about what any of these small refactors mean. I just kept simplifying by equational reasoning, only getting slightly tricky around `refineConfig`, which I'm still reasonably sure is correct.

Having done it, I'm actually unsure whether this is a good idea. Although code is mostly just deleted, it still slightly increases potential for merge conflicts with Ormolu, without much real benefit.

### DIFF
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -184,10 +184,6 @@ formatOne ConfigFileOpts {..} mode reqSourceType rawConfig mpath =
                     <> sourceFile
               return (Just cabalInfo)
             CabalFound cabalInfo -> return (Just cabalInfo)
-        getDotOrmoluForSourceFile' sourceFile = do
-          if optDoNotUseDotOrmolu
-            then return Nothing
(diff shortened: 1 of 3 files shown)


brandonchinn178:
Yes, I avoided this because of the increased potential for merge conflicts. But the conflicts should be relatively easy to resolve, so if you want to merge it, that's fine

georgefst:
Okay. Thinking about it more, I reckon we should go ahead with this.

I don't like the idea of keeping around code related to functionality we don't use at all. And despite Fourmolu's unusual relationship with Ormolu, I think this particular bit of functionality is unique in that respect.

PR #437, opened 2024-10-10
Remove `weeder.toml`

georgefst:


brandonchinn178 (approved):

georgefst:
@brandonchinn178 Any idea what the CI failure is about? I'm not very familiar with our WASM stuff.

brandonchinn178:
Yeah, just ignore it. I was going to take a look at some point in the
future. Presumably Ormolu has it fixed, so it's just a matter of cherry
picking that fix

georgefst:
Okay, cool. I hadn't realised it was optional and already failing on `main`.

PR #458, opened 2025-01-21
Statically link prod binaries

### LINKED ISSUE #457: Provide statically linked binary for easier consumption
**Is your bug specific to Fourmolu?**
- [x] Yes, I've verified that Ormolu does not have this bug

```
[sibi@hask:~/Downloads]$ file ormolu
ormolu: ELF 64-bit LSB executable, x86-64, version 1 (SYSV), statically linked, stripped
```

**Did you try it on the web app?**
- [ ] Yes, I've verified that https://fourmolu.github.io can reproduce this bug

Not applicable

**Did you search for existing issues?**
- [x] Yes, there are no open or closed issues related to my issue

**Describe the bug**
Provide steps to reproduce the bug, and what the expected behavior should be.

Right now, the binary provided in GitHub releases (https://github.com/fourmolu/fourmolu/releases/tag/v0.17.0.0) is dynamically linked which makes me unable to use it in some of the systems that I'm using.

The currently downloaded fourmolu binary is dynamically linked:

```
$ file fourmolu-0.17.0.0-linux-x86_64
fourmolu-0.17.0.0-linux-x86_64: ELF 64-bit LSB executable, x86-64, version 1 (SYSV), dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2, for GNU/Linux 3.2.0, BuildID[sha1]=662c5efb641f1c28932cd9800459f4caa1b9de83, stripped
```

brandonchinn178:
Fixes #457 

Seems like the output size increases by about 50% (~57MB to ~83MB)

### DIFF
--- a/.github/workflows/build-prod-binary.yml
+++ b/.github/workflows/build-prod-binary.yml
@@ -5,6 +5,9 @@ on:
       platforms:
         type: string
         required: true
+      static:
+        type: string
+        required: true
       version:
         type: string
         required: true
@@ -14,6 +17,7 @@ jobs:
     strategy:
       matrix:
         platform: ${{ fromJSON(inputs.platforms) }}
+        static: ${{ fromJSON(inputs.static) }}
 
     name: ${{ matrix.platform }}
     runs-on:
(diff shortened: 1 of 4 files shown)


psibi:
> Seems like the output size increases by about 50% (~57MB to ~83MB)

While this is not a concern for me, if this is issue for some of the consumers - probably worth publishing both static as well as dynamically linked binary ?

Issue #468, opened 2025-04-04
Make 'newlines-between-decls' respectful


ymeister:
**Is your feature request specific to Fourmolu?**
- [x] Yes, this feature is not relevant to Ormolu

**Describe the current + desired formatting**

Original:
```hs

-- Code block 1

data SomeData1 = SomeData1
  { someData1_field1 :: Bool
  , someData1_field2 :: Int
  }

someFunc1_1 :: SomeData1 -> Int
someFunc1_1 = undefined
someFunc1_2 :: SomeData1 -> String
someFunc1_2 = undefined



-- Code block 2

data SomeData2 = SomeData2
  { someData2_field1 :: String
  , someData2_field2 :: Double
  }

someFunc2_1 :: SomeData2 -> Int
someFunc2_1 = undefined

someFunc2_2 :: SomeData2 -> String
someFunc2_2 = undefined
```

Current - `fourmolu --respectful=true --newlines-between-decls=1`:
```hs
-- Code block 1

data SomeData1 = SomeData1
  { someData1_field1 :: Bool
  , someData1_field2 :: Int
  }

someFunc1_1 :: SomeData1 -> Int
someFunc1_1 = undefined

someFunc1_2 :: SomeData1 -> String
someFunc1_2 = undefined

-- Code block 2

data SomeData2 = SomeData2
  { someData2_field1 :: String
  , someData2_field2 :: Double
  }

someFunc2_1 :: SomeData2 -> Int
someFunc2_1 = undefined

someFunc2_2 :: SomeData2 -> String
someFunc2_2 = undefined
```

Desired - `fourmolu --respectful=true --newlines-between-decls=1`:
```hs

-- Code block 1

data SomeData1 = SomeData1
  { someData1_field1 :: Bool
  , someData1_field2 :: Int
  }

someFunc1_1 :: SomeData1 -> Int
someFunc1_1 = undefined

someFunc1_2 :: SomeData1 -> String
someFunc1_2 = undefined



-- Code block 2

data SomeData2 = SomeData2
  { someData2_field1 :: String
  , someData2_field2 :: Double
  }

someFunc2_1 :: SomeData2 -> Int
someFunc2_1 = undefined

someFunc2_2 :: SomeData2 -> String
someFunc2_2 = undefined
```

I'm used to separating code blocks with blank lines, as I find it easier to spot a desired code block when scrolling compared to just trying to spot a comment delimiter.

brandonchinn178:
[...]

PR #469, opened 2025-04-04
Make 'newlines-between-decls' respectful

ymeister:
https://github.com/fourmolu/fourmolu/issues/468


ymeister:
The only thing left is to either amend tests to distinguish between respectful and non-respectful newlines-between-decls or create a new config option that is about respecting blank lines in conjunction with newlines-between-decls.

Issue #467, opened 2025-03-04
K&R brace support


moll:
Hey,

Thanks for maintaining Fourmolu!

Would you be open to supporting/configuring the [K&R brace style](https://en.wikipedia.org/wiki/Indentation_style) instead of the BSD-like that it seems to insist on?

That is, in addition to the default of:

```hs
person = Person
  { name = "John",
    age = 42
  }
```

it'd be possible to go with:

```hs
person = Person {
  name = "John",
  age = 42
}
```

Thanks!

brandonchinn178:
I have no personal issues with it, although FWIW I've seen the vast majority of Haskell programs written in the current style, so I would probably recommend using the current style to be consistent with Haskell style used everywhere. But I have seen a couple projects using this new style, so I'm not totally opposed.

PR welcome!

cblp:
Also, 4-space mode introduces extra 2-space half-indent with current options:
```hs
person =
____Person                -- full indent
    ____{ name = "John",  -- full indent
        __age =           -- half indent
          __age'          -- half indent
        }
```
Good news, this option will get rid of these half-indents:
```hs
person =
____Person {            -- full indent
    ____name = "John",  -- full indent
        age =           
        ____age'        -- full indent
    }
```

kindl:
Hello, I have always used this style and I have seen it in various places, for example in GHC. I implemented a first working version locally, but before doing a PR I wanted to ask:
* Should this be a comma-style or a separate record-style option? I went for a record-style, but I use a similar style for brackets, as well as the combination "KnR with leading comma" is not that sensible. So I could imagine it being a comma-style option.
* I made a specific function recordBraces which no longer uses brackets_, because I need to remove the sitcc at the beginning and I have to de-indent for the closing brace. Is this fine or should I add more options to brackets_?

PR #487, opened 2025-09-18
Add `record-style` option for keeping open brace on the same line as the record constructor (K&R Braces)

kindl:
* Add a setting `record-style`
* The option `aligned` keeps the current behavior, the option `diff-friendly` holds the open brace on the same line as the record constructor
* Requested in #467


brandonchinn178:
I'll review the code in a bit, but I have a gut reaction to the option name. Can you elaborate on why this is diff-friendly? I see this as diff-unfriendly, since it's much more likely to add something to the end of a record than the beginning, and this style would require adding a comma to the previous line

data/fourmolu/record-style/output-diff_friendly-trailing.hs
```diff
@@ -0,0 +1,34 @@
...
+        address :: String,
+        aFieldWithAnUnneccessarilyLongFieldName :: String
+    }
+
+x :: Foo -> [Int]
+x Foo {
+    a = possiblyLongName,
+    b = anotherLongName,
+    c = longNameAsWell
+} = [possiblyLongName, anotherLongName, longNameAsWell]
```
kindl, line 26: I am happy, how the records look as values and data type declarations, but in pattern contexts they look really odd. When formatting the same document a second time, this will look like this:

```
x :: Foo -> [Int]
x
    Foo {
        a = possiblyLongName,
        b = anotherLongName,
        c = longNameAsWell
    } = [possiblyLongName, anotherLongName, longNameAsWell]
```

Which is probably better. Any idea how to achieve this result on the first run, or what is your opinion?

kindl: Probably also:
```
x :: Foo -> [Int]
x Foo {
    a = possiblyLongName
    b = anotherLongName
    c = longNameAsWell
    } = [possiblyLongName, anotherLongName, longNameAsWell]
```

kindl: Good. I have not fully understood, why it renders differently the second time. For now I have added a commit with the second option, but the first option would be more consistent.

kindl: I was still investigating how to get option 1 from the first format run.

The columnLimit confuses me a little. Let's say I set it to 80, then I would assume the following line:

```
x Foo {a = possiblyLongName, b = anotherLongName, c = longNameAsWell} = [possiblyLongName, anotherLongName, longNameAsWell]
```

to break, because it's over 80 characters. But it does not, because the column limit looks at individual items.


In the examples I set it to 60 so that

```
Foo {a = possiblyLongName, b = anotherLongName, c = longNameAsWell}
```

is over the limit and breaks.

Is this intended? If so, the description could be a little bit clearer.

kindl:
Alright, I have hopefully fixed the CI failures.

PR #500, opened 2025-12-31
allow optparse-applicative 0.19

juhp:
This would also unlock fourmolu in Stackage

cf tweag/ormolu#1186


juhp:
And thank you for the Hackage revision!

PR #503, opened 2026-01-14
Apply `indent-wheres` to pattern synonym `where` clauses

### LINKED ISSUE #486: `indent-wheres` is ignored in pattern synonyms (bug?)
- [x] Yes, this feature is not relevant to Ormolu

**Describe the current + desired formatting**
```hs
{-# LANGUAGE PatternSynonyms #-}

-- input
a =
  b
  where
    a = b

pattern A <-
  B
  where
    A = B

-- expected
a =
    b
  where
    a = b

pattern A <-
    B
  where
    A = B

-- got
a =
    b
  where
    a = b

pattern A <-
    B
    where
        A = B
```

emmanuel-ferdman:
## PR Summary
Pattern synonyms with explicit `where` clauses were ignoring the `indent-wheres` setting - they always formatted as if it were enabled. This patch applies the same indentation logic already used for regular functions, so pattern synonyms now behave consistently. With `indent-wheres: false`, the `where` keyword sits at half-indentation; with `indent-wheres: true`, it aligns with the body.

Closes #486.

### DIFF
--- a/src/Ormolu/Printer/Meat/Declaration/Value.hs
+++ b/src/Ormolu/Printer/Meat/Declaration/Value.hs
@@ -1000,9 +1000,13 @@ p_patSynBind PSB {..} = do
               located psb_def p_pat
             breakpoint
             indentWhere <- getPrinterOpt poIndentWheres
-            bool (inciByFrac (-1 / 2)) id indentWhere $ txt "where"
+            let (inciWhere, inciBody) =
+                  if indentWhere
+                    then (id, inci)
+                    else (inciByFrac (-1 / 2), id)
+            inciWhere $ txt "where"
             breakpoint
-            inciIf indentWhere $ p_matchGroup (Function psb_id) mgroup
+            inciBody $ p_matchGroup (Function psb_id) mgroup
   txt "pattern"
   case psb_args of
     PrefixCon [] xs -> do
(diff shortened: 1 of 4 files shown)


src/Ormolu/Printer/Meat/Declaration/Value.hs
```diff
@@ -999,9 +999,10 @@ p_patSynBind PSB {..} = do
               breakpoint
               located psb_def p_pat
             breakpoint
-            txt "where"
+            indentWhere <- getPrinterOpt poIndentWheres
+            bool (inciByFrac (-1 / 2)) id indentWhere $ txt "where"
             breakpoint
-            inci (p_matchGroup (Function psb_id) mgroup)
+            inciIf indentWhere $ p_matchGroup (Function psb_id) mgroup
```
brandonchinn178: Can we do the two `indentWhere` branches once?
```hs
let (inciWhere, inciGroup) = if indentWhere then (id, inci) else (inciByFrac (-1/2), id)
```

brandonchinn178:
@emmanuel-ferdman Can you update the diff files for swarm? I think the breaking change makes sense here (FYI @byorgey). You can get the diff from the CI output: https://github.com/fourmolu/fourmolu/actions/runs/21012217304?pr=503

emmanuel-ferdman:
@brandonchinn178 thanks for the review! Applied all the suggestions 😄

PR #522, opened 2026-04-03
Add `import-list` import grouping option

michivi:
This PR is an attempt to resolve #519: it adds a new `import-list` option to import grouping rules.

The current proposal adds the `import-list` rule option, with the following values:

- `explicit`: matches import declarations with an explicit import list
- `hiding`: matches import declarations with an explicit `hiding` clause
- `none`: matches import declarations without an explicit import list of `hiding` clause (original request from #519)
- no value: matches any import declaration, regardless of presence or absence of an import list or `hiding` clause.

Example configuration:

```yaml
import-grouping:
  - name: "Text modules"
    rules:
      - glob: Data.Text
      - import-list: none
```

While #519 focuses on identifying whole module imports only, this PR also attempts to see what identifying the other cases would look like as well: explicit import lists and `hiding` clauses. Though I'm not sure there are scenarios where the latter would be very useful. So ignoring those might be preferable to keep the code simple... What do you think?

Note that this PR does not change how we interpret `match: all` or local modules as briefly discussed in #519, as it seems out of scope here.

Closes #519

### DIFF
--- a/src/Ormolu/Config/Types.hs
+++ b/src/Ormolu/Config/Types.hs
@@ -51,13 +51,12 @@ instance Aeson.FromJSON ImportGroupRule where
     igrModuleMatcher <- attemptParseModuleMatcher
 
     igrImportListMatcher <-
-      o .:? "import-list" >>= \case
-        Just "any" -> pure MatchAnyImportDeclaration
-        Just "explicit" -> pure MatchExplicitImportList
-        Just "hiding" -> pure MatchHidingImportClause
-        Just "none" -> pure MatchWholeModuleImport
-        Just other -> Aeson.parseFail $ "Unknown import list matcher: " <> other
-        Nothing -> pure MatchAnyImportDeclaration
+      o .:? "import-list" .!= "any" >>= \case
+        "any" -> pure MatchAnyImportDeclaration
+        "explicit" -> pure MatchExplicitImportList
+        "hiding" -> pure MatchHidingImportClause
+        "none" -> pure MatchWholeModuleImport
+        other -> Aeson.parseFail $ "Unknown import list matcher: " <> other
 
     qualified <- o .:? "qualified"
     igrQualifiedMatcher <- case qualified of
--- a/src/Ormolu/Config/Types.hs
+++ b/src/Ormolu/Config/Types.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE RecordWildCards #-}
 
@@ -49,13 +50,13 @@ instance Aeson.FromJSON ImportGroupRule where
         attemptParseModuleMatcher = parseModuleMatcher <|> failUnknownModuleMatcher
     igrModuleMatcher <- attemptParseModuleMatcher
 
-    importList <- Aeson.parseFieldMaybe @String o "import-list"
-    igrImportListMatcher <- case importList of
-      Just "explicit" -> pure MatchExplicitImportList
-      Just "hiding" -> pure MatchHidingImportClause
-      Just "none" -> pure MatchWholeModuleImport
-      Just other -> Aeson.parseFail $ "Unknown import list matcher: " <> other
-      Nothing -> pure MatchAnyImportDeclaration
+    igrImportListMatcher <-
+      o .:? "import-list" >>= \case
+        Just "explicit" -> pure MatchExplicitImportList
+        Just "hiding" -> pure MatchHidingImportClause
+        Just "none" -> pure MatchWholeModuleImport
+        Just other -> Aeson.parseFail $ "Unknown import list matcher: " <> other
+        Nothing -> pure MatchAnyImportDeclaration
 
     qualified <- o .:? "qualified"
     igrQualifiedMatcher <- case qualified of
--- a/src/Ormolu/Config/Types.hs
+++ b/src/Ormolu/Config/Types.hs
@@ -9,6 +9,7 @@ module Ormolu.Config.Types
     matchAllRulePriority,
     matchLocalRulePriority,
     defaultImportRulePriority,
+    ImportListMatcher (..),
     QualifiedImportMatcher (..),
   )
 where
@@ -35,6 +36,7 @@ instance Aeson.FromJSON ImportGroup where
 
 data ImportGroupRule = ImportGroupRule
   { igrModuleMatcher :: !ImportModuleMatcher,
+    igrImportListMatcher :: !ImportListMatcher,
     igrQualifiedMatcher :: !QualifiedImportMatcher,
     igrPriority :: !ImportRulePriority
   }
@@ -46,6 +48,14 @@ instance Aeson.FromJSON ImportGroupRule where
         failUnknownModuleMatcher = Aeson.parseFail "Unknown or invalid module matcher"
         attemptParseModuleMatcher = parseModuleMatcher <|> failUnknownModuleMatcher
     igrModuleMatcher <- attemptParseModuleMatcher
+
+    importList <- Aeson.parseFieldMaybe @String o "import-list"
+    igrImportListMatcher <- case importList of
+      Just "explicit" -> pure MatchExplicitImportList
+      Just "hiding" -> pure MatchHidingImportClause
+      Just "none" -> pure MatchWholeModuleImport
+      Just other -> Aeson.parseFail $ "Unknown import list matcher: " <> other
+      Nothing -> pure MatchAnyImportDeclaration
 
     qualified <- o .:? "qualified"
     igrQualifiedMatcher <- case qualified of
@@ -74,6 +84,13 @@ matchLocalRulePriority = ImportRulePriority 60 -- Lower priority than "all" but 
 
 defaultImportRulePriority :: ImportRulePriority
 defaultImportRulePriority = ImportRulePriority 50
+
+data ImportListMatcher
+  = MatchExplicitImportList
+  | MatchHidingImportClause
+  | MatchWholeModuleImport
+  | MatchAnyImportDeclaration
+  deriving (Eq, Show)
 
 data QualifiedImportMatcher
   = MatchQualifiedOnly
--- a/src/Ormolu/Config/Types.hs
+++ b/src/Ormolu/Config/Types.hs
@@ -52,6 +52,7 @@ instance Aeson.FromJSON ImportGroupRule where
 
     igrImportListMatcher <-
       o .:? "import-list" >>= \case
+        Just "any" -> pure MatchAnyImportDeclaration
         Just "explicit" -> pure MatchExplicitImportList
         Just "hiding" -> pure MatchHidingImportClause
         Just "none" -> pure MatchWholeModuleImport
--- a/tests/Ormolu/Config/PrinterOptsSpec.hs
+++ b/tests/Ormolu/Config/PrinterOptsSpec.hs
@@ -493,76 +493,55 @@ spanEnd f xs =
 importGroupCustomRules :: NonEmpty ImportGroup
 importGroupCustomRules =
   NonEmpty.fromList
-    [ ImportGroup
-        { igName = Nothing,
-          igRules =
-            NonEmpty.fromList
-              [ ImportGroupRule
-                  { igrModuleMatcher = MatchAllModules,
-                    igrImportListMatcher = MatchWholeModuleImport,
-                    igrQualifiedMatcher = MatchUnqualifiedOnly,
-                    igrPriority = defaultImportRulePriority
-                  }
-              ]
-        },
-      ImportGroup
-        { igName = Nothing,
-          igRules =
-            NonEmpty.fromList
-              [ ImportGroupRule
-                  { igrModuleMatcher = MatchGlob (mkGlob "Data.Text"),
-                    igrImportListMatcher = MatchAnyImportDeclaration,
-                    igrQualifiedMatcher = MatchBothQualifiedAndUnqualified,
-                    igrPriority = defaultImportRulePriority
-                  }
-              ]
-        },
-      ImportGroup
-        { igName = Nothing,
-          igRules =
-            NonEmpty.fromList
-              [ ImportGroupRule
-                  { igrModuleMatcher = MatchAllModules,
-                    igrImportListMatcher = MatchAnyImportDeclaration,
-                    igrQualifiedMatcher = MatchBothQualifiedAndUnqualified,
-                    igrPriority = ImportRulePriority 100
-                  }
-              ]
-        },
-      ImportGroup
-        { igName = Nothing,
-          igRules =
-            NonEmpty.fromList
-              [ ImportGroupRule
-                  { igrModuleMatcher = MatchGlob (mkGlob "SomeInternal.**"),
-                    igrImportListMatcher = MatchAnyImportDeclaration,
-                    igrQualifiedMatcher = MatchQualifiedOnly,
-                    igrPriority = defaultImportRulePriority
-                  },
-                ImportGroupRule
-                  { igrModuleMatcher = MatchGlob (mkGlob "Unknown.**"),
-                    igrImportListMatcher = MatchAnyImportDeclaration,
-                    igrQualifiedMatcher = MatchUnqualifiedOnly,
-                    igrPriority = defaultImportRulePriority
-                  }
-              ]
-        },
-      ImportGroup
-        { igName = Nothing,
-          igRules =
-            NonEmpty.fromList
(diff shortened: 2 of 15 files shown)


src/Ormolu/Config/Types.hs
```diff
@@ -47,6 +49,14 @@ instance Aeson.FromJSON ImportGroupRule where
         attemptParseModuleMatcher = parseModuleMatcher <|> failUnknownModuleMatcher
     igrModuleMatcher <- attemptParseModuleMatcher
 
+    importList <- Aeson.parseFieldMaybe @String o "import-list"
+    igrImportListMatcher <- case importList of
+      Just "explicit" -> pure MatchExplicitImportList
+      Just "hiding" -> pure MatchHidingImportClause
+      Just "none" -> pure MatchWholeModuleImport
+      Just other -> Aeson.parseFail $ "Unknown import list matcher: " <> other
+      Nothing -> pure MatchAnyImportDeclaration
```
brandonchinn178: Should be able to simplify:

```suggestion
    igrImportListMatcher <- o .:? "import-list" >>= \case
      Just "explicit" -> pure MatchExplicitImportList
      Just "hiding" -> pure MatchHidingImportClause
      Just "none" -> pure MatchWholeModuleImport
      Just other -> Aeson.parseFail $ "Unknown import list matcher: " <> other
      Nothing -> pure MatchAnyImportDeclaration
```

tests/Ormolu/ConfigSpec.hs
```diff
@@ -8,7 +8,7 @@ import Data.List (isInfixOf)
 import Data.List.NonEmpty qualified as NonEmpty
 import Data.Map qualified as Map
 import Data.Yaml qualified as Yaml
-import Ormolu.Config (FourmoluConfig (..), ImportGroup (..), ImportGroupRule (..), ImportGrouping (..), ImportModuleMatcher (..), ImportRulePriority (ImportRulePriority), PrinterOpts (..), QualifiedImportMatcher (MatchBothQualifiedAndUnqualified, MatchQualifiedOnly, MatchUnqualifiedOnly), defaultImportRulePriority, matchAllRulePriority, resolvePrinterOpts)
+import Ormolu.Config (FourmoluConfig (..), ImportGroup (..), ImportGroupRule (..), ImportGrouping (..), ImportListMatcher (..), ImportModuleMatcher (..), ImportRulePriority (ImportRulePriority), PrinterOpts (..), QualifiedImportMatcher (MatchBothQualifiedAndUnqualified, MatchQualifiedOnly, MatchUnqualifiedOnly), defaultImportRulePriority, matchAllRulePriority, resolvePrinterOpts)
```
brandonchinn178: Can we make this a universal import now? 😅

tests/Ormolu/ConfigSpec.hs
```diff
@@ -82,13 +82,113 @@ spec = do
...
+                          [ ImportGroupRule
+                              { igrModuleMatcher = MatchGlob (mkGlob "**"),
+                                igrImportListMatcher = MatchExplicitImportList,
+                                igrQualifiedMatcher = MatchBothQualifiedAndUnqualified,
+                                igrPriority = defaultImportRulePriority
+                              }
+                          ]
+                    }
+                ]
+        actualStrategy `shouldBe` Just (ImportGroupCustom expectedRules)
```
brandonchinn178: Can we unify all of the `import-list` tests as
```hs
sequence_
  [ it ("enables the '" <> val <> "' import list option") $ do
      ...
  | (val, expected) <- [("explicit", MatchExplicitImportList), ...]
  ]
```
You can unify just the tests with `import-list` present, and leave the last test separate.

Or as a bonus: add an explicit `any` option, and test that `import-list: any` == no `import-list`

tests/Ormolu/ConfigSpec.hs
```diff
@@ -320,6 +432,7 @@ spec = do
                         NonEmpty.fromList
                           [ ImportGroupRule
                               { igrModuleMatcher = MatchGlob (mkGlob "Control.Monad.**"),
+                                igrImportListMatcher = MatchAnyImportDeclaration,
```
brandonchinn178: It should be clear in each test what is actually being tested. Break out a `defaultImportGroupRule` helper with defaults set, and have each test explicitly indicate exactly what is notable in that test

brandonchinn178 (commented):
Some style comments, but core logic looks good!

src/Ormolu/Config/Types.hs
michivi: Sure! Doing so would require adding `LambdaCase`, hence I wasn't sure if it was acceptable. ea9b9e20

tests/Ormolu/ConfigSpec.hs
michivi: Yes! :slightly_smiling_face:  2bb3aef4

tests/Ormolu/ConfigSpec.hs
michivi: I thought about that for readability and maintainability, but it would require to refactor all import grouping tests, and I did not want to either refactor more than was required now (that is, touch other tests which features are untouched in this PR), nor introduce inconsistency (e.g., new tests are explicitly using the helper, while existing tests were not modified). Would you rather we updated all tests in this PR, or delay that to another PR?

tests/Ormolu/ConfigSpec.hs
michivi: 44cf8b9f adds the `any` import list option.

tests/Ormolu/ConfigSpec.hs
michivi: Ok, I started implementing it. We can move the implementation into another PR if we want to (for visibility) :slightly_smiling_face:

- 05f96a12 introduces the helper and refactor the tests introduced for `import-list`
- 80f52146 refactors existing rule attributes tests

src/Ormolu/Config/Types.hs
```diff
@@ -47,6 +50,15 @@ instance Aeson.FromJSON ImportGroupRule where
...
     igrModuleMatcher <- attemptParseModuleMatcher
 
+    igrImportListMatcher <-
+      o .:? "import-list" >>= \case
+        Just "any" -> pure MatchAnyImportDeclaration
+        Just "explicit" -> pure MatchExplicitImportList
+        Just "hiding" -> pure MatchHidingImportClause
+        Just "none" -> pure MatchWholeModuleImport
+        Just other -> Aeson.parseFail $ "Unknown import list matcher: " <> other
+        Nothing -> pure MatchAnyImportDeclaration
```
brandonchinn178: Perhaps do `(o .:? "import-list" .!= "any") >>= \case`, to separate the "parse string" logic from the "what's the default" logic?

tests/Ormolu/Config/PrinterOptsSpec.hs
brandonchinn178: Same here, can we do two things:
1. Break out the custom rules to a top-level definition at the bottom of the file, to avoid cluttering the `TestGroup` list
2. Use a `defaultImportGroupRule`

tests/Ormolu/ConfigSpec.hs
brandonchinn178: Looks great! I was also referring to all of these `expectedRules` definitions; we should have a `defaultImportGroupRule` and then do
```hs
      it "parses a 'match' rule for local modules" $ do
        config <-
          Yaml.decodeThrow . Char8.pack . unlines $
            [ "import-grouping:",
              "  - rules:",
              "      - match: local-modules"
            ]
        let actualStrategy = poImportGrouping (cfgFilePrinterOpts config)
            expectedRules =
              NonEmpty.singleton
                ImportGroup
                  { igName = Nothing,
                    igRules =
                      NonEmpty.singleton
                        defaultImportGroupRule{igrModuleMatcher = MatchLocalModules}
                  }
        actualStrategy `shouldBe` Just (ImportGroupCustom expectedRules)
```

tests/Ormolu/ConfigSpec.hs
```diff
@@ -35,6 +37,38 @@ spec = do
       poIndentation (resolvePrinterOpts configs) `shouldBe` 4
 
     context "when using an import grouping configuration" $ do
+      let checkRuleAttribute desc ruleConfig applyExpectedChange =
```
brandonchinn178: Can we define this further down, colocated with the actual tests?

src/Ormolu/Config/Types.hs
michivi: Right! 9eb79c32

tests/Ormolu/Config/PrinterOptsSpec.hs
michivi: Not entirely sure what you meant with 1. Do you mean moving the definition of the check functions?

9f976fc6

tests/Ormolu/ConfigSpec.hs
michivi: 9f976fc6

tests/Ormolu/Config/PrinterOptsSpec.hs
brandonchinn178: Right now, you have one big list under `ImportGroupCustom`:
```hs
      TestGroup
        { label = "import-grouping",
          isMulti = False,
          testCases =
            [ ImportGroupPreserve,
              ImportGroupSingle,
              ImportGroupByQualified,
              ImportGroupByScope,
              ImportGroupByScopeThenQualified,
              ImportGroupCustom . NonEmpty.fromList $
                <really long list of rules>
            ],
          updateConfig = \igs opts ->
            opts
              { poImportGrouping = pure igs
              },
          showTestCase = \igs ->
            [ renderImportGrouping igs
            ],
          checkIdempotence = True
        },
```
Can you break out that list of rules such that the test case just says
```hs
ImportGroupCustom importGroupCustomRules
```

tests/Ormolu/ConfigSpec.hs
```diff
@@ -338,3 +260,12 @@ spec = do
               Left (Yaml.AesonException msg) -> "Unknown or invalid module matcher" `isInfixOf` msg
               _ -> False
         decodeResult `shouldSatisfy` isAnUnknownModuleMatcher
+
+defaultImportGroupRule :: ImportModuleMatcher -> ImportGroupRule
+defaultImportGroupRule moduleMatcher =
+  ImportGroupRule
+    { igrModuleMatcher = moduleMatcher,
```
brandonchinn178, line 215: This should just be `MatchAllModules`

tests/Ormolu/ConfigSpec.hs
brandonchinn178: Great, I see you broke out `defaultImportGroupRule`. Now you need to change all of these rules from:
```hs
                            ImportGroupRule
                              { igrModuleMatcher = MatchGlob (mkGlob "Control.Monad"),
                                igrImportListMatcher = MatchAnyImportDeclaration,
                                igrQualifiedMatcher = MatchUnqualifiedOnly,
                                igrPriority = defaultImportRulePriority
                              }
```
to
```hs
                            defaultImportGroupRule
                              { igrModuleMatcher = MatchGlob (mkGlob "Control.Monad")
                              }
```

michivi: I originally wasn't sure to update them now, but yeah, it does make sense to do it now given all the last changes. 50eb9e0a

tests/Ormolu/Config/PrinterOptsSpec.hs
michivi: Ah ok! Sorry for the misunderstanding :sweat_smile: 

cf0fdd0e

tests/Ormolu/ConfigSpec.hs
brandonchinn178: ^ This one still

tests/Ormolu/Config/PrinterOptsSpec.hs
```diff
@@ -543,3 +489,80 @@ spanEnd :: (a -> Bool) -> [a] -> ([a], [a])
...
+          igRules =
+            NonEmpty.fromList
+              [ ImportGroupRule
+                  { igrModuleMatcher = MatchAllModules,
+                    igrImportListMatcher = MatchWholeModuleImport,
+                    igrQualifiedMatcher = MatchUnqualifiedOnly,
+                    igrPriority = defaultImportRulePriority
+                  }
+              ]
+        },
```
brandonchinn178: Can we also break out a `defaultImportGroup`/`defaultImportGroupRule` for this? Maybe just in the `where` clause, since it only needs to be used in here.

This way, it'll be easier to see what's different about each group here

tests/Ormolu/ConfigSpec.hs
michivi: Actually, I went with the `moduleMatcher` parameter in `defaultImportGroupRule` as I was afraid having `MatchAllModules` here would be confusing developers into thinking `match` / `glob` is an optional configuration attribute, with `match: all` as a default, which is not the case for the parser. Currently, if neither `match` or `glob` is present, the parser fails telling the user one must be specified.

Would you rather still apply that change? In which case, we may need to rename `default` as it's not really the default value, but rather a testing one. Another alternative would be to update the parser to consider `match: all` as the default. What do you think?

tests/Ormolu/Config/PrinterOptsSpec.hs
michivi: Sure thing 80ae0f6a

tests/Ormolu/Config/PrinterOptsSpec.hs
```diff
@@ -543,3 +489,59 @@ spanEnd :: (a -> Bool) -> [a] -> ([a], [a])
...
+        NonEmpty.fromList
+          [ (defaultImportGroupRule $ MatchGlob (mkGlob "SomeInternal.**"))
+              { igrQualifiedMatcher = MatchQualifiedOnly
+              },
+            (defaultImportGroupRule $ MatchGlob (mkGlob "Unknown.**"))
+              { igrQualifiedMatcher = MatchUnqualifiedOnly
+              }
+          ],
+      defaultImportGroup $
+        NonEmpty.fromList
```
brandonchinn178: Tip: you can reduce one level of nesting with
```hs
defaultImportGroup . NonEmpty.fromList $
  [
  ]
```

tests/Ormolu/ConfigSpec.hs
brandonchinn178: Ah I understand your concern now. You're trying to make this helper mimic the runtime behavior of this configuration; at runtime, it's required to choose either all/local-modules/glob, so you want to require that decision here too. I can sympathize with that, but I still have two counter-points:
1. This is just a test helper, whose only purpose is to make our tests a bit cleaner. It doesn't _have_ to exactly mimic the runtime behavior
2. Following our discussion in the issue, I think the approach we want long-term is to move local-modules to `local-modules: yes|no` and remove `all` in favor of `glob: *` (which would be the default). So even if we don't change that now, we could still write the tests in anticipation of this new approach

But regardless, I think the tests are at a much better state now, so I'll approve it and if you want to merge as-is, I'm happy to do so

brandonchinn178 (approved):

tests/Ormolu/ConfigSpec.hs
michivi: I was thinking about this: even if we rename `defaultImportGroupRule` to something like `baseImportGroupRule` (just in an attempt to remove the potential "is default" confusion), for clarity, I'm guessing it would still be preferable to set the matcher explicitly in the test code. So perhaps we should just keep things as-is for now, and remove the mandatory aspect of the matcher in the parser as soon as possible in a follow-up PR so that the codebase stay overall consistent and clear?

tests/Ormolu/Config/PrinterOptsSpec.hs
michivi: Right 581331ec

brandonchinn178:
@michivi Looks like there's a GHC 9.12 failure

michivi:
> @michivi Looks like there's a GHC 9.12 failure

Yeah, I saw the error, but what seems weird is that it happened during compilation of the `ghc-lib-parser-9.12.1.20250105` dependency, and this PR should not affect that. :thinking:

Plus, previous builds succeeded, and then only change that happened since the last successful build (581331ecd41327adb493eb202d651c304d1dd904) is re-indentation. Only the GHC 9.12 build is affected. I was wondering if we could try and re-run the failing CI job? (I can't restart that specific job)

brandonchinn178:
@michivi Okay should be fixed now. Can you rebase?

michivi:
@brandonchinn178 Thanks, it works now! :slightly_smiling_face:

brandonchinn178:
Thank you for your hard work!

PR #525, opened 2026-04-13
Add `scope` import grouping attribute and deprecate `match`

michivi:
As mentioned in #519, this PR:

- Introduces a `scope` import grouping attribute to determine which import declarations will be matched (external modules, local modules or both). As opposed to what is mentioned in #519, this PR uses `scope: local` rather than `local-module: yes` to improve on flexibility. Options are: `scope: external` to match external modules only, `scope: local` to match local modules only, `scope: any` to match both (by default).
- Deprecates `match`, relying on `glob: **` to replace `match: all` and `scope: local` for `match: local-modules`. Default priorities previously applied by `match` no longer applies to force explicitness. Explicit `scope` and `glob` take precedence over `match`.


michivi:
Sorry, forgot to update the changelog :sweat_smile: 5e56fc15

brandonchinn178 (approved):
Beautiful, thank you!

Issue #456, opened 2025-01-20
Per-file {- FOURMOLU_OPTIONS -} pragma


AshleyYakeley:
Feature suggestion:

Allow per-file specification of options with a `FOURMOLU_OPTIONS` pragma. Syntax would match the command-line. For example:

```
{- FOURMOLU_OPTIONS --import-grouping preserve -}
```

bfeitknecht:
this seems like a great addition!

brandonchinn178:
Seems doable! Should be able to modify `cfgWithIndices` here using the parsing result in `result0`:
https://github.com/fourmolu/fourmolu/blob/9b131d7a056d4a969386e594907e393943b3035e/src/Ormolu.hs#L121

PR welcome!

PR #527, opened 2026-04-30
Add local Fourmolu options

michivi:
This PR is an attempt at #456, adding a `{- FOURMOLU_OPTIONS <local-config> -}` comment to customize the Fourmolu formatting configuration for a single file.

Note that the syntax is not using `{-# ... #-}` to prevent the `Unrecognised pragma: FOURMOLU_OPTIONS` warning.

The first commit attempts to use the Fourmolu CLI options for the `<local-config>` placeholder. But while it is something usually done by most tools (e.g., `{-# OPTIONS_GHC ... #-}`), implementing that for Fourmolu is quite a pain:

- The CLI parsing logic is in `Main`, from the main executable. It is not available to the `Ormolu` module responsible for the formatting. Refactoring this would make it harder to keep Fourmolu in sync with Ormolu, and could potentially break projects using the library (e.g., the Haskell Language server)
- While we could try and use `parsePrinterOptsCLI`, we would still have to add `optparse-applicative` as a dependency to the main library which may not be desirable. Not relying on it means adding parsing code that adds a lot of complexity that may not desirable as well.

The second commit (the current PR proposal) represents the local configuration as YAML. While this diverges from what most tools do (i.e., provide CLI arguments), it would on the other hand improve expressiveness (e.g., more complex import grouping configuration).

### DIFF
--- a/src/Ormolu/Exception.hs
+++ b/src/Ormolu/Exception.hs
@@ -45,6 +45,8 @@ data OrmoluException
     OrmoluMissingStdinInputFile
   | -- | A parse error in a fixity overrides file
     OrmoluFixityOverridesParseError (ParseErrorBundle Text Void)
+  | -- | A FOURMOLU_OPTIONS pragma could not be parsed in the original source code
+    FourmoluPragmaOptsParsingFailed RealSrcSpan Text
   deriving (Show)
 
 instance Exception OrmoluException where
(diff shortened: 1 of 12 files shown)


src/Ormolu/Exception.hs
```diff
@@ -136,3 +146,4 @@ withPrettyOrmoluExceptions colorMode m = m `catch` h
           OrmoluCabalFileParsingFailed {} -> 8
           OrmoluMissingStdinInputFile {} -> 9
           OrmoluFixityOverridesParseError {} -> 10
+          FourmoluPragmaOptsParsingFailed {} -> 16
```
michivi, line 149: Making this further apart from potentially new exit code from Ormolu.

PR #538, opened 2026-07-11
Allow empty config file

kafji:
Currently there are 3 solutions to configure Fourmolu to use its default options:

1. By ensuring no configuration files in its lookup paths

2. By creating a configuration file from `fourmolu --print-defaults`

3. By creating a configuration file with at least one option e.g. `indentation: 4`

The first solution will fail if there's a configuration file somewhere, e.g. when `~/.config/fourmolu.yaml` exists but project's `./fourmolu.yaml` does not, the 2nd and the 3rd implies the project uses non default options.

This patch expands the solutions with creating an empty configuration file.

I'm particularly interested in this because the haskell-language-server test for Fourmolu will fail if the host machine has non default options in their `~/.config/fourmolu.yaml`.

### DIFF
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -18,6 +18,7 @@ import Data.Maybe (fromMaybe, mapMaybe, maybeToList)
 import Data.Set qualified as Set
 import Data.Text.IO.Utf8 qualified as T.Utf8
 import Data.Version (showVersion)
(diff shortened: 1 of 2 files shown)


brandonchinn178:
Technically, you could create a file with `{}`, would that work?

kafji:
It does! Thank you.

PR #542, opened 2026-08-07
Merge ormolu 0.8.2.0

brandonchinn178:


compat-tests/cabal.diff
brandonchinn178, line 1: Big compat diff with the fixity changes. Not much we can do about this, since it's Ormolu making the breaking change

